PR #80, opened 2026-04-01
Refactor(validation): Rebuild validation around the Trait ValidatableValue

ClausHolbechArista:
## Change Summary

Rebuild validation around the Trait ValidatableValue

## Component(s) name

rust

## Proposed changes

- Preparation for yaml-parser. This code is mainly split out from that bigger PR to make it more reviewable.
- Abstracting the data value type for validation behind a Trait that we can implement for different types like JSON, YAML and maybe Pyo3 types.
- Moving coercion to happen during validation instead of two parses.

## How to test

Existing testing passes. Added new tests as well where applicable.

### User Checklist

- N/A

### Repository Checklist


rust/validation/src/validatable/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    /// Get the type of this value for error reporting.
+    fn value_type(&self) -> Type;
+
+    /// Convert to feedback::Value for use in CoercionNote.
+    /// This is used to record what value was coerced from.
+    fn to_feedback_value(&self) -> crate::feedback::Value;
+
+    /// Returns true if this value is a float.
+    fn is_float(&self) -> bool;
```
gmuloc: should be moved with other `is_foo`

rust/validation/src/validation/boolean.rs
```diff
@@ -3,46 +3,64 @@
...
 
 use super::Validation;
 
-impl Validation<bool> for Bool {
-    fn validate(&self, value: &bool, ctx: &mut Context) {
-        self.validate_ref(value, ctx);
-    }
+impl Validation for Bool {
+    fn validate<V: ValidatableValue>(&self, value: &V, ctx: &mut Context) -> Option<V::Coerced> {
+        if let Some(ref_result) = validate_ref(self, value, ctx) {
```
gmuloc: a bit pedantic but technically ref_result is an Option and not a Result

rust/validation/src/validation/boolean.rs
```diff
@@ -3,46 +3,64 @@
...
+            ctx.add_error_for(
+                value,
+                Violation::InvalidType {
+                    expected: Type::Bool,
+                    found: value.value_type(),
+                },
+            );
+            None
         }
     }
```
gmuloc: we could abstract this handling in the `ctx` with two helpers

maybe_coerce_null and then ctx.add_type_error or something to avoid repeating again and again the couple of lines for every schemna?

rust/validation/src/validation/dict.rs
```diff
@@ -31,60 +30,84 @@ const EOS_CLI_CONFIG_GEN_ROLE_KEYS: [&str; 8] = [
...
+/// structured_config, where we $ref in the full eos_config schema.
+fn validate_ref<V: ValidatableValue>(
+    schema: &Dict,
+    value: &V,
+    ctx: &mut Context,
+) -> Option<Option<V::Coerced>> {
+    if let Some(ref_) = schema.base.schema_ref.as_ref() {
+        // Ignoring not being able to resolve the schema.
+        // Ignoring a wrong schema type at the ref. Since Validation is infallible.
+        // TODO: What to do?
```
gmuloc: these comments were removed in bool but kept here - was it on purpose?

rust/validation/src/validation/dict.rs
```diff
@@ -31,60 +30,84 @@ const EOS_CLI_CONFIG_GEN_ROLE_KEYS: [&str; 8] = [
...
+    schema: &Dict,
+    input: &M,
+    ctx: &mut Context,
+) -> Option<Vec<(String, <M::Value as ValidatableValue>::Coerced)>> {
+    let return_coerced = ctx.configuration.return_coerced_data;
+    let mut coerced_items = if return_coerced {
+        Some(Vec::new())
+    } else {
+        None
+    };
```
gmuloc:
```suggestion
    let mut coerced_items = ctx.configuration.return_coerced_data.then(Vec::new);
```

rust/validation/src/validation/dict.rs
```diff
@@ -31,60 +30,84 @@ const EOS_CLI_CONFIG_GEN_ROLE_KEYS: [&str; 8] = [
...
-            dynamic_keys
-                .into_iter()
-                .map(|(key, dynamic_key_info)| (key, dynamic_key_info.schema))
-                .collect()
-        })
-        .unwrap_or_default()
-}
+/// Validate and optionally coerce mapping keys.
+/// Returns Some(coerced_items) when coercion is enabled, None otherwise.
+fn validate_keys<'a, M: ValidatableMapping<'a>>(
```
gmuloc, line 85: this method is quite big and doing a lot of things

rust/validation/src/validation/dict.rs
```diff
@@ -137,52 +206,66 @@ fn validate_required_keys(schema: &Dict, input: &Map<String, Value>, ctx: &mut C
...
     if let Some(deprecation) = key_schema.deprecation()
         && deprecation.warning
     {
         if deprecation.removed.unwrap_or_default() {
-            ctx.add_error(Violation::Removed(Removed::from_schema(
-                &ctx.state.path,
-                deprecation,
-            )));
+            ctx.add_error_with_span(
+                key_span,
```
gmuloc, line 225: I know this compiles but not sure why we don't need to clone key_span here when we need for warning so leaving a comment to review later, especially when we clone it later

rust/validation/src/validation/dict.rs
```diff
@@ -557,28 +686,57 @@ mod tests {
...
     }
 
+    /// Test that keys starting with underscore are preserved in output but not validated
+    #[test]
+    fn validate_underscore_key_preserved() {
+        let schema = Dict {
+            keys: Some(OrderMap::from_iter([("foo".into(), Str::default().into())])),
+            ..Default::default()
+        };
+        // _internal key should be preserved but not validated, foo1 should error
```
gmuloc: there is no foo1

rust/validation/src/validation/dict.rs
```diff
@@ -604,13 +764,16 @@ mod tests {
                 })
             }]
         );
+        // With lenient validation, int 123 coerces to string "123"
```
gmuloc: what is lenient validation?

rust/validation/src/validation/list.rs
```diff
@@ -4,70 +4,166 @@
...
+    schema: &List,
+    input: &S,
+    ctx: &mut Context,
+) -> Option<Vec<<S::Value as ValidatableValue>::Coerced>> {
+    let return_coerced = ctx.configuration.return_coerced_data;
+    let mut coerced = if return_coerced {
+        Some(Vec::with_capacity(input.len()))
+    } else {
+        None
+    };
```
gmuloc:
```suggestion
let mut coerced = ctx.configuration.return_coerced_data.then(|| Vec::with_capacity(input.len()));
```

rust/validation/src/validation/list.rs
```diff
@@ -4,70 +4,166 @@
...
+
+    for (i, item) in input.iter().enumerate() {
+        ctx.state.path.push(i.to_string());
+        validate_item_primary_key(schema, item, ctx);
+        if let Some(ref mut items) = coerced {
+            let coerced_item = validate_item_schema(schema, item, ctx);
+            items.push(coerced_item);
+        } else {
+            validate_item_schema_only(schema, item, ctx);
         }
```
gmuloc, line 77: I feel we could rewrite this to have only one validate_item function that returns an Option<V::Coerced> depending on the return_coerced_data value

rust/validation/src/validation/store.rs
```diff
@@ -3,159 +3,175 @@
...
-        &self,
-        value: &mut Value,
-        schema_name: &str,
-    ) -> Result<ValidationResult, StoreValidateError>;
+    ) -> Result<ValidationOutput<V::Coerced>, StoreValidateError>;
 }
 
-impl StoreValidate for Store {
+/// Parse JSON input and validate the resulting value against a schema.
+pub trait StoreValidateInput {
```
gmuloc, line 49: I am not too sure about why splitting the trait in two helps. I find the two names a bit confusing. Any reason?

rust/validation/src/validation/store.rs
```diff
@@ -3,159 +3,175 @@
...
 }
 
 #[derive(Debug, derive_more::Display, derive_more::From)]
 pub enum StoreValidateError {
     SchemaStore(avdschema::SchemaStoreError),
-    Json(serde_json::Error),
-    Yaml(serde_yaml::Error),
+}
+
+fn empty_validation_output<T>() -> ValidationOutput<T> {
```
gmuloc: could we not implement a `#[derive(Default)]` on the ValidationOutput? Given that we have default on the validationResult I think it would just work

rust/validation/src/validation/boolean.rs
```diff
@@ -3,46 +3,64 @@
...
-        self.validate_ref(value, ctx);
-    }
+impl Validation for Bool {
+    fn validate<V: ValidatableValue>(&self, value: &V, ctx: &mut Context) -> Option<V::Coerced> {
+        if let Some(ref_result) = validate_ref(self, value, ctx) {
+            return ref_result;
+        }
 
-    fn validate_value(&self, value: &Value, ctx: &mut Context) {
         if let Some(v) = value.as_bool() {
```
gmuloc, line 17: not emitting coercion info for bool?

rust/validation/src/validation/str.rs
```diff
@@ -3,103 +3,184 @@
...
 
-    fn validate_ref(&self, value: &String, ctx: &mut Context) {
-        if let Some(ref_) = self.base.schema_ref.as_ref() {
-            // Ignoring not being able to resolve the schema.
-            // Ignoring a wrong schema type at the ref. Since Validation is infallible.
-            // TODO: What to do?
-            if let Ok(AnySchema::Str(ref_schema)) = resolve_ref(ref_, ctx.store) {
-                ref_schema.validate(value, ctx);
+/// Emit coercion info if the original value was coerced to string.
+fn emit_coercion_info<V: ValidatableValue>(value: &V, coerced_str: &str, ctx: &mut Context) {
```
gmuloc: maybe this could be moved to ctx by passing it the right info and checking return_coercion_infos inside the main validate?

rust/validation/src/validation/str.rs
```diff
@@ -3,103 +3,184 @@
...
+                    },
+                );
             }
+            lower
+        } else {
+            s
         }
+    } else {
+        s
     }
```
gmuloc, line 60:
```suggestion
    if !schema.convert_to_lower_case.unwrap_or_default() {
        return s;
    }

    let lower = s.to_lowercase();

    if lower != s {
       if ctx.configuration.return_coercion_infos {
            ctx.add_info_for(
                value,
                CoercionNote {
                    found: s.into(),
                    made: lower.clone().into(),
                },
            );
        }
        return lower;
    }
    s
}
```

rust/validation/src/validation/str.rs
```diff
@@ -3,103 +3,184 @@
...
+fn convert_to_lower_case<V: ValidatableValue>(
+    schema: &Str,
+    value: &V,
+    s: String,
+    ctx: &mut Context,
+) -> String {
+    if schema.convert_to_lower_case.unwrap_or_default() {
+        let lower = s.to_lowercase();
+        if lower != s {
+            if ctx.configuration.return_coercion_infos {
```
gmuloc: if we build ctx.maybe_emit_coercion_info we can probably simplify this in several places (avoid the ctx.config + ctx.add repeat everywhere)

rust/validation/src/validation/str.rs
```diff
@@ -3,103 +3,184 @@
...
+            s
         }
+    } else {
+        s
     }
 }
 
-fn validate_min_length(schema: &Str, input: &str, ctx: &mut Context) {
+/// Validate against a referenced schema (for unresolved $ref ending with #).
+fn validate_ref<V: ValidatableValue>(
```
gmuloc, line 64: I feel like this could be abstracted in some way for all types rather than rewriting almost the same across all validation enum value.

rust/validation/src/validation/dict.rs
ClausHolbechArista: cleaned up

rust/validation/src/validation/dict.rs
ClausHolbechArista: to avoid iterating over the keys in multiple places we need to centralize it. Let's wait for the stack of PRs to get in before deciding. I have abstracted the logic of keys vs dynamic kys vs prefix keys so it will simplify this function.

rust/validation/src/validation/list.rs
ClausHolbechArista: there is a lot of cloning happening that we would like to avoid when we don't need it, and the items.push is also expensive to to when it is not needed.

rust/validation/src/validation/dict.rs
ClausHolbechArista: just that we coerce wrong types

rust/validation/src/validation/dict.rs
ClausHolbechArista: in the case of this `if` we don't need key_span again, so we _move_ it in this call.
In the `else` it is needed early and again multiple times inside a loop so we have to clone both places. The spans are fairly cheap to clone.

rust/validation/src/context.rs
```diff
@@ -90,6 +126,11 @@ pub struct Configuration {
     /// When validating avd_design, emit warnings for top-level keys that exist in eos_config
     /// but not in avd_design.
     pub warn_eos_config_keys: bool,
+    /// When true, validation returns coerced data with types adjusted according to the schema.
+    /// When false (default), validation returns a null placeholder to avoid expensive cloning.
+    /// Set to true when you need the coerced output (e.g., for data transformation).
+    /// Set to false for validation-only use cases (e.g., LSP diagnostics).
+    pub return_coerced_data: bool,
```
gmuloc: I would group this with the other `return_coercion_infos` for clarity

rust/validation/src/context.rs
```diff
@@ -26,42 +29,75 @@ impl<'a> Context<'a> {
             state: Default::default(),
         }
     }
-    pub(crate) fn add_error(&mut self, error: impl Into<ErrorIssue>) {
+    pub(crate) fn add_error_for<V: ValidatableValue>(
```
gmuloc, line 35: is it really necessary to keep both version (_for and _with_span) when we could just pass the value span instead of calling add_error_for?

rust/validation/src/feedback.rs
gmuloc: this description seems outdated

rust/validation/src/feedback.rs
```diff
@@ -153,14 +160,102 @@ pub enum InfoIssue {
...
+/// Absolute byte range within the original source input.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct SourceSpan {
+    pub start: usize,
+    pub end: usize,
+}
+
+/// Parse diagnostic reported while decoding structured input.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct ParseDiagnostic {
```
gmuloc, line 190: I feel this PR is doing two separate things: adding the trait and adding the diagnostic bit. This makes the review a bit harder but I will review both at the same time.

I am a bit confused about the Input Diagnstic being in parallel of the other feedbacks so maybe we need to discuss this live

rust/validation/src/feedback.rs
```diff
@@ -153,14 +160,102 @@ pub enum InfoIssue {
...
+}
+
+/// Parse diagnostic category.
+#[derive(Clone, Debug, PartialEq, Serialize, derive_more::Display)]
+pub enum ParseDiagnosticKind {
+    #[display("JSON syntax error")]
+    JsonSyntax,
+}
+
+fn source_span_from_json_error(input: &str, error: &serde_json::Error) -> SourceSpan {
```
gmuloc, line 237: probably worth a oneliner comment explaining what this does

rust/validation/src/feedback.rs
```diff
@@ -153,14 +160,102 @@ pub enum InfoIssue {
...
+            None => String::new(),
+        }
+    }
+}
+
+impl Display for ParseDiagnostic {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}: {}.",
```
gmuloc, line 219: if the error message contains a final `.` we may end up with `..` though it seems that won't happen for serde_json so for now we are good.

rust/validation/src/feedback.rs
```diff
@@ -153,14 +160,102 @@ pub enum InfoIssue {
...
 #[derive(Clone, Debug, PartialEq, Serialize, derive_more::Display)]
 #[display("Coerced value from {found} to {made}.")]
 pub struct CoercionNote {
     pub found: Value,
     pub made: Value,
 }
 
+/// Absolute byte range within the original source input.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct SourceSpan {
```
gmuloc, line 183: do we then convert this back to something? how it is used in presentation?

rust/validation/src/validation/list.rs
gmuloc: yeah agreed hence the option right?

rust/validation/src/feedback.rs
ClausHolbechArista: This is getting changes again in a later PR, to have a SourceInfo enum which is either line/column or bytes depending on what we get from the underlying parser lib.

rust/validation/src/feedback.rs
ClausHolbechArista: yeah best practice overall for error messages is to not capitalize and not add a final punctiation, but we are mostly doing the opposite everywhere, so I kept taht style here.

rust/validation/src/feedback.rs
ClausHolbechArista: you marked a lot, are you only talking about the source_span_from_json_error? I think that is pretty clear by itself, but again this will change in a [subsequent PR](https://github.com/ClausHolbechArista/pyavd-utils/pull/5) that is already in the stack.

rust/validation/src/validation/store.rs
ClausHolbechArista: The type is external (owned by avdschema) so we cannot implement the flat functions with hardcoded types directly on the struct. That is the reason for the StoreValidateInput.
Then we have the StoreValidate<V> with is abstracted depending on the type of value.

rust/validation/src/validation/list.rs
ClausHolbechArista: the only other option is to wrap the extra things done in validate_item_schema helper function up here into the if/else to both can validate on the items schema directly, but it will not make it more readable.

rust/validation/src/validation/str.rs
ClausHolbechArista: IMO context should not carry _logic_ in itself.

ClausHolbechArista: Ah maybe I misunderstood. But storing the messages just to throw them away is not very efficient. They will almost never be used.

rust/validation/src/validation/boolean.rs
ClausHolbechArista: We are not coercing _into_ a bool. Check implementation. Maybe we need to make this clearer with `.coerce_as_<>` on the ones where we support it.

rust/validation/src/context.rs
ClausHolbechArista: Changing the order may change behavior if someone was not setting them with named args.

rust/validation/src/validation/str.rs
ClausHolbechArista: yes agreed such a helper would make sense.

rust/validation/src/validation/str.rs
ClausHolbechArista: I will build a helper on ctx.

rust/validation/src/context.rs
ClausHolbechArista: just convenience. We may want to add other contextual information later.

rust/validation/src/validation/str.rs
ClausHolbechArista: We unpack the result of resolve_ref differently for each. I don't think it makes sense to spend more time here at the moment, since we have talked about finding better ways of expanding the $refs ad hoc.

rust/validation/src/context.rs
ClausHolbechArista: i was wrong. Moving.

rust/validation/src/feedback.rs
ClausHolbechArista: It is not strictly needed now, since we still only have JSON, but when I add yaml-parser I will be creating parse errors as well as validation errors, so this new wrapper is really necessary there. Could have been a separate PR.

rust/validation/src/validation/store.rs
ClausHolbechArista: done

rust/validation/src/validation/store.rs
ClausHolbechArista: We can discuss how we want the outer API to look and redesign it seperately. Today we build the store and then call validate on it, but we can also build clean functions instead and take the store as an argument?

rust/validation/src/validation/str.rs
ClausHolbechArista: So I modified this and now I have a test breakage where `Bool(true)` becomes `true` because I now add the original Bool(True) as the "from" value in this coercion where before I had `String(True)->true`. 

Before:
- `Bool(true)->Str("True")` and `Str("True")->Str("true")`

After:
- `Bool(true)->Str("True")` and `Bool(true)->Str("true")`

But you could also argue that we should only do one coercion info so:
- `Bool(true)->Str("true")`

ClausHolbechArista: For now changing back so the test pass

ClausHolbechArista: I cannot use the helper because I need to emit the intermediate value which is not the same type as elsewhere where we call the helper, so reverted this callpoint.

rust/validation/src/validation/list.rs
gmuloc: I feel like we could merge the two helpers together with a boolean flag saying if we want back the coerced item or not that would avoid extra cloning like

```
fn validate_item(schema, item, ctx, return_coerced) -> Option<V::Coerced>:

```

and then take decisions inside this function based on the bool and push only in the main one.

though I will resolve for now :)

rust/validation/src/feedback.rs
gmuloc: discussed on our call - lets proceed

rust/validation/src/feedback.rs
gmuloc: discussed on the call - resolving

rust/validation/src/feedback.rs
gmuloc: yeah it wasa for the source_span_from_json_error. we will update it later.

rust/validation/src/validatable/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// - Float → String (e.g., `3.14` → `"3.14"`)
+    /// - Bool → String (e.g., `true` → `"true"`)
+    ///
+    /// Returns `Cow::Borrowed` if already a string, `Cow::Owned` if coerced.
+    fn as_str(&self) -> Option<Cow<'_, str>>;
+
+    /// Try to get this value as a 64-bit signed integer, coercing if possible.
+    ///
+    /// Coerces:
+    /// - String → Int (if parseable, e.g., `"123"` → `123`)
```
gmuloc, line 79: This is incomplete as the implementation for serde_json also convert bool -> Int and float -> int. We should update the description for future contract

rust/validation/src/validation/list.rs
```diff
@@ -77,64 +173,65 @@ fn validate_unique_keys(schema: &List, items: &[Value], ctx: &mut Context) {
...
-            key: primary_key.to_owned(),
-        });
+    if let Some(b) = value.as_bool() {
+        return b.to_string();
     }
+    if value.is_null() {
+        return "null".to_string();
+    }
+    // For complex types, we can't easily compare
+    "<complex>".to_string()
```
gmuloc: this should be a TODO right? as this look like something that can bite us back one day

rust/validation/src/validation/store.rs
gmuloc: ok understood the limitation and the reason why - good to go for now on tyhis one

rust/validation/src/validation/boolean.rs
gmuloc: yeah the implementation is only for json for now but indeed it got me confused as I cant know if we would coerce for other Validatable - it may be worth making the distinction

rust/validation/src/validatable/mod.rs
ClausHolbechArista: Done. Good catch. This went a bit back and forth so missed the update here.

rust/validation/src/validation/list.rs
ClausHolbechArista: it is only used in unique_keys comparisons and indexing. Not even for the error message. We don't use unique_keys to compare larger constructs, only "leaf" keys.

ClausHolbechArista: looking at this again, the coercion would succeed on the as_str() for both ints and bools so I can remove those two. And "null" may also overlab with someone giving the string `"null"` so I will add some __ before and after

rust/validation/src/validation/boolean.rs
ClausHolbechArista: I have made a plan for renaming we can discuss, but I will postpone it until the inflight work has merged.

gmuloc: ack seen the plan - resolving this

rust/validation/src/validation/list.rs
gmuloc: > it is only used in unique_keys comparisons and indexing. Not even for the error message. We don't use unique_keys to compare larger constructs, only "leaf" keys.

I know today we don't in our schema (same for primar_keys) but that could be something fragile for the future. Anyways for now we can carry on :)

rust/validation/src/context.rs
```diff
@@ -137,15 +155,15 @@ pub(crate) struct State {
...
-    pub return_coerced_data: bool,
     /// By default Null/None values are ignored no matter which data type is expected.
     /// Setting this will instead emit type errors for Null values.
-    pub return_coercion_infos: bool,
+    pub restrict_null_values: bool,
     /// When true, validation returns coerced data with types adjusted according to the schema.
     /// When false (default), validation returns a null placeholder to avoid expensive cloning.
+    pub return_coerced_data: bool,
+    /// When validating avd_design, emit warnings for top-level keys that exist in eos_config
+    /// but not in avd_design.
```
gmuloc: this is in the wrong place

ClausHolbechArista: Not sure how I can get this wrong in so many ways :)

gmuloc (approved):

PR #82, opened 2026-04-01
Refactor(rust): Abstract data type for avdschema helpers

ClausHolbechArista:
## Change Summary

Abstract data type for avdschema helpers

## Proposed changes

To avoid tying the logic of the AVD schema directly to serde_json::Value, we abstract the input types for the exposed helper functions to be defined with a Trait. This will still only be used for serde_json::Value, but allows us to extend this nicely in the validation trait instead of implementing everything again in the validation code.

## How to test

Existing tests are covering everything we touch here.

### User Checklist

- N/A

### Repository Checklist


carl-baillargeon (approved):
LGTM if CI passes.

PR #85, opened 2026-04-12
Feat(rust): Add yaml-parser crate

ClausHolbechArista:
## Change Summary

Add yaml-parser crate.

This is not being used yet, but was taken out of a bigger PR to make reviews simpler.

## Component(s) name

rust

## Proposed changes

`yaml-parser` is a YAML 1.2 parser with three main goals:

- recover from syntax and structure errors without stopping at the first issue
- preserve source spans throughout the pipeline
- keep the hot path allocation-light and zero-copy where possible

The crate exposes two main user-facing parsing styles:

- AST parsing via [`parse`](src/lib.rs)
- serde deserialization via [`yaml_parser::serde`](src/serde/mod.rs)

Both share the same lexer and emitter.

See ARCHITECTURE.md and BENCHMARKS.md for details.

## How to test

- Extensive unit and integration testing, including testing the official YAML test suite with 100% compliance.
- including scripts to run remote benches on a stable bare metal system.
- profiling script (under examples) to do deep dive performance optimizations.

### User Checklist

- N/A

### Repository Checklist


Cargo.toml
gmuloc: any reason to not make `rust/yaml-parser` a default-members? It means that some of the CI we are running may not run. e.g clippy:

https://github.com/aristanetworks/pyavd-utils/actions/runs/24769464000/job/72471867317?pr=85#step:7:15

gmuloc: and worse, it looks like test are not runnign: https://github.com/aristanetworks/pyavd-utils/actions/runs/24769464000/job/72471867317?pr=85#step:8:15

ClausHolbechArista: happened because of parallel PRs. I will update.

ClausHolbechArista: Fixing in https://github.com/aristanetworks/pyavd-utils/pull/95. I have already pulled those change into this one so it should run the tests properly now.

rust/yaml-parser/src/lexer/directives.rs
```diff
@@ -0,0 +1,180 @@
...
+impl<'input> Lexer<'input> {
+    /// Check if we're at column 0 (start of input or right after a newline).
+    pub(super) fn is_at_column_zero(&self) -> bool {
+        if self.byte_pos == 0 {
+            return true;
+        }
+        // Check if previous byte was a newline
+        let prev_byte_pos = self.byte_pos.saturating_sub(1);
+        if let Some(ch) = self.input.as_bytes().get(prev_byte_pos) {
+            *ch == b'\n' || *ch == b'\r'
```
gmuloc, line 21: why not use is_newline here (there could be corner case)

ClausHolbechArista: is_newline takes a char, but here we grab a byte so we do a local check without having to decode utf-8 and reindexing the input string to find the last char.

rust/yaml-parser/src/lexer/properties.rs
```diff
@@ -0,0 +1,178 @@
...
+        )
+    }
+
+    pub(super) fn consume_tag(&mut self, start: usize) -> Spanned<Token<'input>> {
+        let tag_start = start; // Position of the '!'
+        self.advance(); // consume !
+
+        // Check for !! or !<
+        match self.peek() {
+            Some('!') => {
```
gmuloc: feels like this case is the same as the one for is_valida_tag_start_char except we specifically consume the ! first

rust/yaml-parser/src/lexer/quoted.rs
```diff
@@ -0,0 +1,345 @@
...
+                    break;
+                }
+            }
+        }
+        if let Ok(code) = u32::from_str_radix(&hex, 16)
+            && let Some(ch) = char::from_u32(code)
+        {
+            return ch.to_string();
+        }
+        format!("\\x{hex}") // Invalid escape - keep as-is
```
gmuloc: should we not record some kind of error for this? (not sure on this one)

rust/yaml-parser/src/lexer/trivia.rs
```diff
@@ -0,0 +1,105 @@
...
+            self.advance();
+            let span = self.current_span(start);
+            self.add_error(ErrorKind::InvalidComment, span);
+            // Try to recover by consuming rest of line as if it were a comment
+            while let Some(peek_ch) = self.peek() {
+                if Self::is_newline(peek_ch) {
+                    break;
+                }
+                self.advance();
+            }
```
gmuloc: this pattern is there in many places - wondering if it would be worth adding a consume_line or whatever helper

rust/yaml-parser/src/lexer/trivia.rs
```diff
@@ -0,0 +1,105 @@
...
+        &mut self,
+        start: usize,
+        ch: char,
+    ) -> Option<Spanned<Token<'input>>> {
+        if ch != '#' {
+            return None;
+        }
+
+        if !self.prev_was_separator {
+            // `#` without preceding whitespace is invalid - report error
```
gmuloc, line 72: I am unsure about this one - I feel like

```
key: blah#666
```

should be interpreted as 

```
<plain: "key"> <column> <space> <plain: "blah#666">
```

given: https://yaml.org/spec/1.2.2/#66-comments 

Comments must be [separated](https://yaml.org/spec/1.2.2/#separation-spaces) from other tokens by [white space](https://yaml.org/spec/1.2.2/#white-space-characters) characters.

SO I would intepret this as if there is no previous separator then it is not a comment  but the spec is but blurry on this so I think we can go on with thsi and take a bug later if it turns out to eb wrong

rust/yaml-parser/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+mod error;
+mod event;
+mod lexer;
+mod parser;
+mod span;
+mod stream;
+mod value;
+
+#[cfg(feature = "serde")]
+mod ast_events;
```
gmuloc: this looks a bit sneaky to have singular vs plural gated by feature - need to understand

rust/yaml-parser/BENCHMARKS.md
```diff
@@ -0,0 +1,132 @@
...
+- Relative to `saphyr_marked`, `yaml_parser` is ahead on `medium` and `large`
+  parse latency, but still behind on `small`.
+- Relative to `serde_yaml`, `yaml_parser` remains ahead on all 7/7
+  serde-deserialize throughput datasets.
+- The scalar microbenchmarks are still all behind `saphyr_marked`, with the
+  largest gap in `block_scalars`.
+
+## Parse Throughput
+
+Median throughput from the remote candidate report.
```
gmuloc: median without confidence interval makes it hard to estimate the proper results here - from what I understand criterion would give also the statisticalin tervals and given the emphasis we put on performance it would be better than just median numbers in our claims I think

gmuloc (commented):
so in terms of comments today we are dropping everycomment that is not after a key it seems (from some locla tests I have tried with parse)

rust/yaml-parser/src/span.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+impl SourceMap {
+    /// Create a new source map from input text.
+    #[must_use]
+    pub fn new(input: &str) -> Self {
+        let mut line_starts = vec![0];
+
+        for (byte_pos, ch) in input.char_indices() {
+            if ch == '\n' {
```
gmuloc: I think this would render incorrect data for `\r` only end of line

rust/yaml-parser/src/lexer/properties.rs
ClausHolbechArista: this is a fast path for the more common case.

ClausHolbechArista: added a comment

rust/yaml-parser/src/value.rs
```diff
@@ -0,0 +1,761 @@
...
+/// node properties, not value wrappers. For example, in `&anchor key: value`,
+/// the anchor attaches to the scalar `key`, which is then used as a mapping key.
+///
+/// The lifetime `'input` refers to the input string being parsed. String content
+/// uses `Cow<'input, str>` for zero-copy when possible.
+///
+/// Properties (anchor, tag) are boxed so nodes without properties stay small,
+/// trading an occasional heap allocation for a more compact base node layout.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Node<'input> {
```
gmuloc, line 309: in this one we are having the properties and comment coming from event - could this make the API not stable if event and Node properties ever need to diverge?

rust/yaml-parser/src/lib.rs
ClausHolbechArista: looks like it is the wrong feature as well, so yeah I will take a look.

rust/yaml-parser/src/lexer/trivia.rs
ClausHolbechArista: The blah#666 would have been picked up by `consume_plain_scalar` already so we only get here for other cases. I don't have a case right now but I'm sure there is test somewhere :)

rust/yaml-parser/src/value.rs
ClausHolbechArista: Yeah I did not see a need to redefine the same thing and this way I can carry over the data without allocating a new struct.

rust/yaml-parser/BENCHMARKS.md
ClausHolbechArista: the target has been to get soo little noise that this did not really matter, but yes I will expand the table here.

rust/yaml-parser/src/lexer/quoted.rs
ClausHolbechArista: invalid escape is just a string

rust/yaml-parser/src/span.rs
ClausHolbechArista: Fixed and also added testing both here and in integration tests.

rust/yaml-parser/src/lexer/properties.rs
ClausHolbechArista: I will simplify.

rust/yaml-parser/src/tests.rs
```diff
@@ -0,0 +1,1443 @@
...
+        }
+        .expect("expected mapping");
+        assert_eq!(pairs.len(), 3, "Should recover all mapping pairs");
+    }
+
+    /// Test that a malformed flow mapping entry without a colon is dropped
+    /// instead of recovered as a fake `null` pair.
+    #[test]
+    fn test_missing_colon_in_flow_mapping_drops_pair_and_recovers() {
+        let input = "{\"a\" [1], c: 3}";
```
gmuloc, line 566: so here the parser is recovering this as

```
{
  a - Invalid  pair,
  [1]: , # empty value
  c: 3
}
```

From the code this is to be expected but it looks "unaligned" with the test description so just wanted to point that out

rust/yaml-parser/src/lexer/quoted.rs
ClausHolbechArista: rewrote the escape handling to produce proper errors.

rust/yaml-parser/src/writer.rs
```diff
@@ -0,0 +1,789 @@
...
+                    self.write_node()?;
+                }
+                Event::DocumentEnd { .. } => {
+                    // End of the current document. For now we do not emit an
+                    // explicit `...` marker, but we ensure we advance past the
+                    // event and terminate the current line.
+                    // If the document had an explicit end marker in the
+                    // original stream, preserve that here so that
+                    // `DocumentEnd { explicit: true, .. }` roundtrips.
+                    if let Some(Event::DocumentEnd { explicit, .. }) = self.peek().cloned() {
```
gmuloc: I am a bit struggling with this part where we need to double check the DocumentEnd - it feals like we are doing double the work. I am being told it is to protect in case for some reason (unclear to me) the cursor as moved since we started the match. Though I dont see how this would happen today.

That being said it works so I am fine resolving this comment and moving on

rust/yaml-parser/src/lexer/trivia.rs
ClausHolbechArista: added advance_until_newline().

This pattern was not exactly the same in many places. Most places are doing something inside the loop.

rust/yaml-parser/src/writer.rs
```diff
@@ -0,0 +1,789 @@
...
+            buf.push_str(&name);
+            self.out.write_all(buf.as_bytes())?;
+            self.at_line_start = false;
+        }
+        Ok(())
+    }
+
+    fn write_indent(&mut self) -> io::Result<()> {
+        for _ in 0..self.indent {
+            self.out.write_all(b" ")?;
```
gmuloc: this feels unoptimized if we have a lot of spaces to write? (maybe I am missing some point)

apparently we could use a buffer or chunked single-write approach.

rust/yaml-parser/src/lib.rs
ClausHolbechArista: renamed it to make it more clear and also cleaned up the wrong comment inside.

rust/yaml-parser/src/tests.rs
ClausHolbechArista: this was changed since the test was written. I have updated the description here. Thanks.

rust/yaml-parser/src/writer.rs
ClausHolbechArista: The writer is mainly used for testing at the moment, so it has not received a lot of attention. I will try to clean this up.

ClausHolbechArista: fixed

rust/yaml-parser/src/writer.rs
ClausHolbechArista: optimized with a chunk approach. Up to 64 spaces per chunk.

rust/yaml-parser/BENCHMARKS.md
ClausHolbechArista: added confidence info above the table to avoid cluttering too much.

rust/yaml-parser/src/lexer/trivia.rs
gmuloc: yeah I tested this will resolve now

rust/yaml-parser/src/emitter/mod.rs
```diff
@@ -0,0 +1,997 @@
...
+
+    /// Take collected errors from both lexer and emitter.
+    pub(crate) fn take_errors(&mut self) -> Vec<ParseError> {
+        let mut all = self.cursor.take_lexer_errors();
+        all.extend(std::mem::take(&mut self.errors));
+        all
+    }
+
+    fn set_pending_event(&mut self, event: Event<'input>) {
+        debug_assert!(
```
gmuloc, line 231: I have read about debug_assert and understand it wont make it to the prod code compiled with `--release` but are we sure we want this as this could hide rewriting of pending_event in prod code that would be hard to understand?

rust/yaml-parser/src/emitter/value.rs
```diff
@@ -0,0 +1,1171 @@
...
+                // Defer alias handling to the `AliasValue` state so that
+                // complex-key behaviour is driven by the state machine.
+                let Some((alias_name, alias_span)) = self
+                    .peek_with(|tok, span| match tok {
+                        Token::Alias(name) => Some((Cow::Borrowed(*name), span)),
+                        _ => None,
+                    })
+                    .flatten()
+                else {
+                    debug_assert!(false, "expected Alias token");
```
gmuloc, line 771: same question for debug_assert - would it not make our life harder to tshoot in the field?

rust/yaml-parser/src/emitter/value.rs
```diff
@@ -0,0 +1,1171 @@
...
+        }
+
+        // Now check what follows the properties.
+        // If it's LineStart, this line is "property-only" - continue collecting.
+        // If it's content, check if that content is an implicit key (followed by colon).
+        match self.peek_kind_nth(idx) {
+            Some(TokenKind::FlowSeqStart | TokenKind::FlowMapStart) => {
+                // Flow collection - check if it's a complex key
+                // Look for matching close bracket, then colon
+                let is_key = self.is_implicit_key_at_offset(idx);
```
gmuloc: `is_implicit_key_at_offset` is called only here and so it looks like 

But I tested and still we parse things correctly just FlowMapStart matching is not really detecting the nest mapping as key?

It shoud also look at quoted strings and single quoted string as 


```
    let input = "!!map\n!foo {a: b}: c\n"; <<< DuplicateTag
    let input = "!!map\n!foo [a, b]: c\n"; <<<< foo on Sequence, no error
     let input = "!!map\n!foo \"a, b\": c\n"; <<<< DuplicateTag


```

rust/yaml-parser/src/emitter/value.rs
```diff
@@ -0,0 +1,1171 @@
...
+                    {
+                        result.push(char::from(byte));
+                        continue;
+                    }
+                    // Failed to decode, keep as-is
+                    result.push('%');
+                    result.push_str(&hex);
+                } else {
+                    result.push(ch);
+                }
```
gmuloc: this can fail for more complex encoding:

```
let input = "!caf%C3%A9 1";
```

should be `café` for the tag

```
[rust/yaml-parser/src/tests.rs:1497:5] &errors = []
[rust/yaml-parser/src/tests.rs:1498:5] &events = [
    StreamStart,
    DocumentStart {
        explicit: false,
        span: Span {
            start: 0,
            end: 10,
        },
    },
    Scalar {
        style: Plain,
        value: "1",
        properties: Some(
            Properties {
                anchor: None,
                tag: Some(
                    Property {
                        value: "!cafÃ©",
                        span: Span {
                            start: 0,
                            end: 10,
                        },
                    },
                ),
            },
        ),
        span: Span {
            start: 11,
            end: 12,
        },
    },
    DocumentEnd {
        explicit: false,
        span: Span {
            start: 12,
            end: 12,
        },
    },
    StreamEnd,
]
+STR
+DOC
=VAL <!cafÃ©> :1
-DOC
-STR
```

rust/yaml-parser/src/emitter/value.rs
```diff
@@ -0,0 +1,1171 @@
...
+        })
+    }
+
+    /// Expand a tag handle to its full form.
+    ///
+    /// The lexer transforms tags as follows:
+    /// - `!!str` → `Tag("!str")` (secondary handle)
+    /// - `!name!suffix` → `Tag("name!suffix")` (named handle)
+    /// - `!<uri>` → `Tag("\0uri")` (verbatim, marked with NUL)
+    /// - `!` alone → `Tag("")` (non-specific)
```
gmuloc: is it still correct? It looks like we now borrow the initial `!`

rust/yaml-parser/src/emitter/mod.rs
ClausHolbechArista: Alternative would be panic!?

rust/yaml-parser/src/emitter/value.rs
ClausHolbechArista: Rebuilt with a better decoder and added tests.

rust/yaml-parser/src/emitter/value.rs
ClausHolbechArista: updated. It was indeed misleading and contradicting the inner comments.

rust/yaml-parser/src/emitter/mod.rs
ClausHolbechArista: added a comment after analyzing alternatives.

rust/yaml-parser/src/emitter/value.rs
ClausHolbechArista: We dispatch on the lightweight `TokenKind` first because extracting or matching on the full token on every path is measurably more expensive.
If `peek_kind()` says `Alias`, the backing token must be `Token::Alias`; a mismatch here indicates an internal desynchronization bug, so we check it with `debug_assert!`.
Basically we rely on the testing to expose bugs.

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+        // Normal path: get from emitter
+        let event = self.emitter.next()?;
+
+        // Check if this is an Alias - if so, start replaying
+        if let Event::Alias { ref name, .. } = event {
+            let name_str = name.to_string();
+            if let Some(recorded) = self.anchors.get(&name_str) {
+                // Clear any peeked event to avoid confusion during replay
+                self.peeked = None;
```
gmuloc: for this nullifying peeked - I was concerned it could "drop" some event at some point but given the way we are calling things peeked should always be None if I understand correctly - in that case would it not be better to raise if we have peeked set and we are here? (or below)

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+    /// Clear all stored anchors.
+    ///
+    /// This should be called between documents to ensure anchors don't leak
+    /// across document boundaries.
+    pub(crate) fn clear_anchors(&mut self) {
+        self.anchors.clear();
+    }
+
+    fn invalid_pair_error() -> DeError {
+        DeError::Custom("invalid mapping entry: expected ':' after the key".into())
```
gmuloc, line 272: is this the only kind of error we can get for invalidaPair?

**EDIT**: yes today but we are coupling the error message for InvalidPair to missing column when we also have an event MissingColumn, I guess thats ok for today.

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+    match text.as_bytes().first() {
+        None => return ScalarKind::Null,
+        Some(b'~') if text.len() == 1 => return ScalarKind::Null,
+        Some(b'n') if text == "null" => return ScalarKind::Null,
+        Some(b'N') if matches!(text, "Null" | "NULL") => return ScalarKind::Null,
+        Some(b't') if text == "true" => return ScalarKind::Bool(true),
+        Some(b'T') if matches!(text, "True" | "TRUE") => return ScalarKind::Bool(true),
+        Some(b'f') if text == "false" => return ScalarKind::Bool(false),
+        Some(b'F') if matches!(text, "False" | "FALSE") => return ScalarKind::Bool(false),
+        Some(b'0'..=b'9' | b'+' | b'-' | b'.') => {
```
gmuloc: to support YAML 1.2 spec we should also have hexdecimal and octal support https://yaml.org/spec/1.2.2/#24-tags

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+        return NumericKind::NotNumeric;
+    }
+    if saw_dot || saw_exp {
+        NumericKind::Float
+    } else {
+        NumericKind::Integer
+    }
+}
+
+fn infer_scalar_kind(value: Cow<'_, str>) -> ScalarKind<'_> {
```
gmuloc: My friend is telling me, the inconsistency is:

  - parse("inf") -> string
  - serde::from_str::<f64>("inf") -> float inf

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+            // Consume the null and report `None`.
+            self.next_event();
+            visitor.visit_none()
+        } else {
+            // Not a null scalar – let the nested deserializer consume it
+            visitor.visit_some(self)
+        }
+    }
+
+    fn deserialize_enum<V>(
```
gmuloc, line 533: this method is failing on round trips for Enums:


```rust
 #[test]
  fn debug_enum_yaml_inputs() {
      use serde::{Deserialize, Serialize};

      #[derive(Debug, Serialize, Deserialize, PartialEq)]
      enum Demo {
          Unit,
          Newtype(i64),
          Struct { x: i64 },
      }

      let inputs = [
          ("unit_plain", "Unit\n"),
          ("unit_quoted", "\"Unit\"\n"),
          ("newtype_mapping", "\"Newtype\": 42\n"),
          ("struct_mapping", "\"Struct\":\n  \"x\": 7\n"),
          ("newtype_scalar_only", "Newtype\n"),
          ("struct_scalar_only", "Struct\n"),
      ];

      for (name, yaml) in inputs {
          eprintln!("=== {name} ===");
          eprintln!("INPUT:\n{yaml}");

          let decoded = crate::serde::from_str::<Demo>(yaml);
          eprintln!("DECODED: {decoded:?}");

          if let Ok(value) = decoded {
              let encoded = crate::serde::to_string(&value);
              eprintln!("RE-ENCODED: {encoded:?}");
          }

          eprintln!();
      }

      panic!("inspect output");
  }
```


```
=== unit_plain ===
  INPUT:
  Unit

  DECODED: Ok(Unit)
  RE-ENCODED: Ok("\"Unit\"\n")

  === unit_quoted ===
  INPUT:
  "Unit"

  DECODED: Ok(Unit)
  RE-ENCODED: Ok("\"Unit\"\n")

  === newtype_mapping ===
  INPUT:
  "Newtype": 42

  DECODED: Err(ExpectedEnumString)

  === struct_mapping ===
  INPUT:
  "Struct":
    "x": 7

  DECODED: Err(ExpectedEnumString)

  === newtype_scalar_only ===
  INPUT:
  Newtype

  DECODED: Err(Custom("invalid type: unit variant, expected newtype variant"))

  === struct_scalar_only ===
  INPUT:
  Struct

  DECODED: Err(Custom("invalid type: unit variant, expected struct variant"))
```

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: For parser we have the same: https://github.com/ClausHolbechArista/pyavd-utils/blob/feat/rust/yaml-parser-crate/rust/yaml-parser/src/parser/scalars.rs#L172

It is `".inf"`, not `"inf"`

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: It is another case of "we know this should never happen", so yes we could panic in that case.

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+/// Parse a YAML 1.2 core schema boolean literal (case-insensitive variants of
+/// `true` / `false`). This mirrors the bool handling in `infer_scalar_kind`
+/// but is cheap enough to use in the hot paths of primitive `deserialize_*`
+/// methods without running full scalar inference.
+#[inline]
+fn parse_core_bool(input: &str) -> Option<bool> {
+    // Fast path: check first byte and length to avoid full string comparison
+    match (input.len(), input.as_bytes().first()?) {
+        (4, b't' | b'T') => input.eq_ignore_ascii_case("true").then_some(true),
```
gmuloc: this would accept `TrUe` and `FaLsE` and so the crate would behave differently if I use the main entrypoint `parse` or if I use crate::serde::from_str which is weird

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+    fn deserialize_f32<V>(self, visitor: V) -> Result<V::Value, DeError>
+    where
+        V: Visitor<'de>,
+    {
+        // Delegate to f64 and let serde handle the narrowing conversion.
+        self.deserialize_f64(visitor)
+    }
+
+    fn deserialize_f64<V>(self, visitor: V) -> Result<V::Value, DeError>
```
gmuloc, line 695: this can sometimes yield different result when deserializing big interger into f64 which would be lossy

`1234567890123456789012345678901234567890`

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: moved to a shared scalar_resolver.rs which is used by parser and serde and enhanced it to support everything required by the "Core schema" in YAML spec 1.2.2

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+        match event {
+            Event::Scalar { style, value, .. } => {
+                if style != ScalarStyle::Plain {
+                    return Err(DeError::Custom(format!(
+                        "invalid f64 value: {}",
+                        value.as_ref()
+                    )));
+                }
+                let scalar = value.as_ref();
+                if let Ok(float) = scalar.parse::<f64>() {
```
gmuloc: this will accept `inf`, `nan` and `infinity` which is different than what we want in YAML spec and so we may end up with unexpected round trips maybe

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: Yeah there are different ways of getting to a missing colon error, and this one is specifically when we only have weak signs during emitting so we have to start a mapping pair but then invalidate it when we figure out it was not correct. I would like to work more on this later to avoid some lookaheads and so on, but it will require substantial changes.

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: I am using debug_assert for consistency. But with the current code this cannot happen. So it is more to catch a posible issue during later changes.

rust/yaml-parser/src/serde/event_de.rs
```diff
@@ -0,0 +1,1261 @@
...
+        let event = self.next_value_event()?;
+        match event {
+            Event::Scalar { style, value, .. } => {
+                EventStream::deserialize_scalar(style, value, visitor)
+            }
+            _ => Err(DeError::Custom("expected scalar for bytes".into())),
+        }
+    }
+
+    fn deserialize_byte_buf<V>(self, visitor: V) -> Result<V::Value, DeError>
```
gmuloc, line 789: this can cause trouble with its serde pendant for 
```
      #[derive(Debug, PartialEq)]
      struct NeedsByteBuf(Vec<u8>);
```

where


```
      let value = NeedsByteBuf(vec![1, 2, 3]);
      let yaml = crate::serde::to_string(&value).unwrap();
      let decoded = crate::serde::from_str::<NeedsByteBuf>(&yaml);
```

```
  [rust/yaml-parser/src/tests.rs:XXXX:5] &yaml = "- 1\n- 2\n- 3\n"
  [rust/yaml-parser/src/tests.rs:XXXX:5] &decoded = Err(
      Custom(
          "expected scalar for bytes",
      ),
  )
```

rust/yaml-parser/src/serde/ser.rs
```diff
@@ -0,0 +1,622 @@
...
+        let key_value = self.next_key.take().ok_or(SerError::ValueWithoutKey)?;
+        let key_node = node_from_value(key_value);
+        let val_node = node_from_value(value.serialize(ValueSerializer)?);
+        self.entries
+            .push(MappingPair::new(val_node.span, key_node, val_node));
+        Ok(())
+    }
+
+    fn end(self) -> Result<Self::Ok, Self::Error> {
+        Ok(Value::Mapping(self.entries))
```
gmuloc, line 412: here should check thast next_key is not set here when called (it could happen if the calling serializer call serialize_key then end)

rust/yaml-parser/src/writer.rs
```diff
@@ -0,0 +1,786 @@
...
+                            self.advance();
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn write_block_sequence(&mut self, sequence_props: &Properties<'input>) -> io::Result<()> {
```
gmuloc, line 470: this will cause issue for empty sequences in serde (same for mappings)

rust/yaml-parser/src/serde/ser.rs
```diff
@@ -0,0 +1,622 @@
...
+{
+    let root_value = ValueSerializer::to_value(value)?;
+    let node = node_from_value(root_value);
+    let events = ast_to_events::node_to_events(&node)?;
+    writer::write_yaml_from_events(&mut writer, &events)?;
+    Ok(())
+}
+
+/// Serialize any `T: Serialize` to a YAML string.
+pub fn to_string<T>(value: &T) -> Result<String, SerError>
```
gmuloc, line 500: yaml_parser::serde::to_string(&Vec::<i32>::new())
  // => Ok("")

  yaml_parser::serde::from_str::<Vec<i32>>("")
  // => Err(NoDocument)

  yaml_parser::serde::to_string(&BTreeMap::<String, i32>::new())
  // => Ok("")

  yaml_parser::serde::from_str::<BTreeMap<String, i32>>("")
  // => Err(NoDocument)

  For a nested empty sequence field:

  #[derive(Debug, Serialize, Deserialize, PartialEq)]
  struct EmptySeqField {
      items: Vec<i32>,
  }

  yaml_parser::serde::to_string(&EmptySeqField { items: vec![] })
  // => Ok("\"items\":\n")

  yaml_parser::serde::from_str::<EmptySeqField>("\"items\":\n")
  // => Err(Custom("expected sequence start for seq deserialization"))


this goes with the comment in writer.rs that we are missing representation of empty collections

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: fixed in new scalar_resolver.rs

rust/yaml-parser/src/parser/stream.rs
```diff
@@ -0,0 +1,165 @@
...
+                AstEvent::Event(Event::StreamStart) => {
+                    // Skip the stream start marker.
+                    self.advance();
+                }
+                AstEvent::Event(Event::StreamEnd) => {
+                    // Consume the stream end marker and signal EOF.
+                    self.advance();
+                    return None;
+                }
+                AstEvent::Event(Event::DocumentStart { .. })
```
gmuloc, line 66: from a maintenance point of view having to match on AstEvent::Event and AstEvent::RichEvent again and again feels a bit of a burden. I understand that the RichEvent is used to attach extra stuff to the Event but this could lead to drift in the long run (if we are not careful)

rust/yaml-parser/src/serde/ser.rs
ClausHolbechArista: since entries is not populated until the value has also ben serialized we don't include the entry if they do that half-baked call series.

rust/yaml-parser/src/ast_event.rs
```diff
@@ -0,0 +1,100 @@
...
+        pair_start: BytePosition,
+        key_event: Event<'input>,
+        /// Comment captured on the structural header line before the wrapped
+        /// key event starts.
+        ///
+        /// This is usually `None` for ordinary scalar keys, but it can be set
+        /// when the wrapped event is a collection start from an implicit
+        /// mapping. In that case it flows to the resulting `ParsedNode` as
+        /// leading comment metadata and may become the parent pair's
+        /// `MappingPair.header_comment`.
```
gmuloc, line 80: need to double check if this may actually happen

gmuloc:
```
  ? # key header
    [1, 2]
  : value
```

in this case the comment is gone and not attached to the key

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: added a todo to improve this

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: EDIT: ~that is just how floats work. Not sure what you want me to do here?~

I agree the important question is where to draw the line between coercion and error. Serde’s built-in f64 behavior is already mixed here: u64 can be accepted lossily, while u128 is rejected. I’d rather make the YAML deserializer policy explicit and consistent instead of inheriting that boundary accidentally from Serde internals. My preference is either:

- permissive numeric-to-f64 coercion, accepting the usual floating-point precision loss, or
- strict behavior where non-float YAML scalars are rejected for f64.

I don’t think a partly-lossy / partly-error policy based on integer width is the right rule unless we intentionally want that.

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: fixed in new scalar_resolver.rs

rust/yaml-parser/src/emitter/value.rs
ClausHolbechArista: this has been fixed for all types and including tests.

rust/yaml-parser/src/writer.rs
ClausHolbechArista: added todo

rust/yaml-parser/src/ast_event.rs
ClausHolbechArista: the description says it will be put on the parent pairs header comment. is that not the case?

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: For the deserialize_<type> I will add todos. General inference has been moved to the common scalar_resolver which handles everything consistently.

rust/yaml-parser/src/serde/ser.rs
ClausHolbechArista: added todo

rust/yaml-parser/src/parser/anchors.rs
```diff
@@ -0,0 +1,64 @@
...
+    I::Item: Into<AstEvent<'input>>,
+{
+    /// Build a resolved alias node.
+    pub(super) fn build_alias(
+        &mut self,
+        name: &Cow<'input, str>,
+        span: Span,
+    ) -> Option<Node<'input>> {
+        if let Some(node) = self.anchor_nodes.get(name.as_ref()) {
+            return Some(node.clone().into_resolved_alias_root(span));
```
gmuloc, line 29: This clone could become very MemoryExpensive for aliases used 1000 time in a document. Apparently using Rc<> (ref count) could be a solution

rust/yaml-parser/src/parser/anchors.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        if self.anchors.contains(name.as_ref()) {
```
gmuloc, line 32: so this (sadly) valid YAML - would break here:

```
&parent_anchor
name: "Big Boss"
employee_list:
  - name: "Small Fry"
    reports_to: *parent_anchor  # <--- ERROR HERE

Comment (gmuloc) [reply rust/yaml-parser/src/parser/anchors.rs:32]: https://yaml.org/spec/1.2.2/#321-representation-graph this states we can have cycles .. maybe we call this out as a limitation? I have seen other parser doing it. (preact)

File: rust/yaml-parser/src/parser/scalars.rs [hunk 34]
```diff
@@ -0,0 +1,203 @@
...
+impl<'input, I> Parser<'input, I>
+where
+    I: Iterator,
+    I::Item: Into<AstEvent<'input>>,
+{
+    /// Build a scalar node with type inference.
+    ///
+    /// Plain scalars are inferred before explicit tags are interpreted as
+    /// annotations, so `!!str 42` currently becomes `Int(42)` with a `!!str`
+    /// tag attached to the node.
```
gmuloc: this seems like potentially non trivial maybe it should trickle to some public facing documentation in the limitations given that it feels we essentially ignore tags?

rust/yaml-parser/src/parser/scalars.rs
```diff
@@ -0,0 +1,203 @@
...
+        let typed_value = if style == ScalarStyle::Plain {
+            infer_scalar_type(value)
+        } else {
+            // Quoted/block scalars are always strings
+            Value::String(value)
+        };
+
+        let base_node = Node::new(typed_value, span);
+        let node = Self::apply_properties(base_node, props);
+        self.store_anchor_node(&node);
```
gmuloc, line 62: If I follow here we can sometimes save a node that is raising below invalidValue - so every alias using this anchor will represent a node that was deemed invalid but not know it (though it may make a lot of error messages to capture this as well)

rust/yaml-parser/src/parser/scalars.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        node
+    }
+}
+
+/// Classify whether a scalar is a plausible integer, a plausible float, or
+/// definitely not numeric using a single byte scan.
+#[inline]
+fn classify_numeric(input: &str) -> NumericKind {
```
gmuloc: same comment as somewhere else on octal and hex

rust/yaml-parser/src/serde/ser.rs
ClausHolbechArista: added todo

rust/yaml-parser/src/parser/anchors.rs
ClausHolbechArista: Alternative is to not resolve the Alias but keep it as a reference type to the achor node. I tried at first that but it gave other issues later when validating etc. So I decided not to support the cycles, since we need to flatten the data.

rust/yaml-parser/src/parser/scalars.rs
ClausHolbechArista: the comment is old. I changed this just Yesterday with the new scalar_resolver.rs

rust/yaml-parser/src/parser/scalars.rs
ClausHolbechArista: fixed as part of the scalar_resolver.rs

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: added todo

rust/yaml-parser/src/parser/anchors.rs
ClausHolbechArista: added todo

rust/yaml-parser/src/parser/anchors.rs
ClausHolbechArista: added a limitations section in architecture.md

rust/yaml-parser/src/parser/scalars.rs
ClausHolbechArista: I don't follow the comment, but the code changed a bit now so check again please.

rust/yaml-parser/src/parser/stream.rs
ClausHolbechArista: yeah I felt the same but rigorous benching/testing said this is the best pattern.

ClausHolbechArista: I have shared a plan to improve comment handling so this will be changed a bit.

rust/yaml-parser/src/value.rs
ClausHolbechArista: shared a suggested plan to address this and stop carrying the anchor in the AST nodes.

rust/yaml-parser/src/ast_event.rs
ClausHolbechArista: This will be solved by a complete comment handling overhall. I shared a plan.

rust/yaml-parser/src/serde/ser.rs
ClausHolbechArista: added error handling.

rust/yaml-parser/src/serde/event_de.rs
ClausHolbechArista: moved discussion to limitations

ClausHolbechArista: moved discussion to limitations in architecture.md

rust/yaml-parser/src/scalar_resolver.rs
gmuloc, line 1: Three things that are inconsistent today:


```
extern crate yaml_parser;
use yaml_parser::parse;

fn main() {
      for input in [
          "!!float 42",
          "!!null ~",
          "-0x80000000000000000000000000000000",
      ] {
          let (docs, errors) = parse(input);
          println!("INPUT: {input}");
          println!("ERRORS: {errors:?}");
          println!("DOC: {:?}\n", docs.first().map(|d| &d.value));
      }
}
```
outputs

```
  INPUT: !!float 42
  ERRORS: [ParseError { kind: InvalidValue, span: Span { start: 8, end: 10 } }]
  DOC: Some(String("42"))

  INPUT: !!null ~
  ERRORS: [ParseError { kind: InvalidValue, span: Span { start: 7, end: 8 } }]
  DOC: Some(String("~"))

  INPUT: -0x80000000000000000000000000000000
  ERRORS: []
  DOC: Some(Int(BigIntStr("-170141183460469231731687303715884105728")))
```


  - !!float 42: currently rejected, recovered as string.
  - !!null ~: currently rejected, recovered as string.
  - -0x8000...: currently accepted, but stored as BigIntStr instead of a concrete I128.


(As comparison)

```python
>>> import yaml
>>> print(repr(yaml.safe_load('!!float 42'))
... )
42.0
>>> print(repr(yaml.safe_load('!!null ~')))
None
>>> print(repr(yaml.safe_load('-0x80000000000000000000000000000000')))
-170141183460469231731687303715884105728
>>>
```

ClausHolbechArista: fixed

PR #89, opened 2026-04-14
Refactor(rust): Replace regex crate with fancy-regex to support lookaround patterns

Shivani-gslab:
## Change Summary

Replace `regex` carte with `fancy-regex` to support lookaround patterns in schema.

## Related Issue(s)

Fixes #<ISSUE ID>

## Proposed changes

- Use fancy-regex for pattern matching in avdschema and validation
- Added tests for the same

## How to test
Added tests for lookaround patterns

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/rust/avdschema/src/resolve/errors.rs
+++ b/rust/avdschema/src/resolve/errors.rs
@@ -13,7 +13,7 @@ use super::walker::SchemaWalkError;
 pub enum SchemaResolverError {
     SchemaType(SchemaType),
     RefSyntax(RefSyntax),
-    RefRegex(RefRegexError),
+    RefRegex(RefRegex),
     SchemaPath(SchemaPath),
     SchemaStoreError(SchemaStoreError),
     SchemaWalkError(SchemaWalkError),
@@ -35,7 +35,7 @@ pub struct RefSyntax {
 
 #[derive(Debug, derive_more::Constructor, derive_more::Display)]
 #[display("Regex error for schema $ref '{schema_ref}': {regex_error}.")]
-pub struct RefRegexError {
+pub struct RefRegex {
     pub schema_ref: String,
     pub regex_error: String,
 }
--- a/rust/avdschema/src/resolve/errors.rs
+++ b/rust/avdschema/src/resolve/errors.rs
@@ -13,6 +13,7 @@ use super::walker::SchemaWalkError;
 pub enum SchemaResolverError {
     SchemaType(SchemaType),
     RefSyntax(RefSyntax),
+    RefRegexError(RefRegexError),
     SchemaPath(SchemaPath),
     SchemaStoreError(SchemaStoreError),
     SchemaWalkError(SchemaWalkError),
@@ -30,6 +31,13 @@ pub struct SchemaType {
 #[display("Invalid syntax for schema $ref '{schema_ref}'.")]
 pub struct RefSyntax {
     pub schema_ref: String,
+}
+
+#[derive(Debug, derive_more::Constructor, derive_more::Display)]
+#[display("Regex error for schema $ref '{schema_ref}': {regex_error}.")]
+pub struct RefRegexError {
+    pub schema_ref: String,
+    pub regex_error: String,
 }
 
 #[derive(Debug, derive_more::Constructor, derive_more::Display)]
--- a/rust/avdschema/src/resolve/errors.rs
+++ b/rust/avdschema/src/resolve/errors.rs
@@ -13,7 +13,7 @@ use super::walker::SchemaWalkError;
 pub enum SchemaResolverError {
     SchemaType(SchemaType),
     RefSyntax(RefSyntax),
-    RefRegexError(RefRegexError),
+    RefRegex(RefRegexError),
     SchemaPath(SchemaPath),
     SchemaStoreError(SchemaStoreError),
     SchemaWalkError(SchemaWalkError),
--- a/rust/avdschema/src/resolve/errors.rs
+++ b/rust/avdschema/src/resolve/errors.rs
@@ -13,7 +13,6 @@ use super::walker::SchemaWalkError;
 pub enum SchemaResolverError {
     SchemaType(SchemaType),
     RefSyntax(RefSyntax),
-    RefRegex(RefRegex),
     SchemaPath(SchemaPath),
     SchemaStoreError(SchemaStoreError),
     SchemaWalkError(SchemaWalkError),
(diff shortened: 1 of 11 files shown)


rust/avdschema/src/schema/any.rs
```diff
@@ -12,12 +10,14 @@ use crate::{
...
     utils::{
         dump::Dump,
-        load::{Load, LoadError},
+        load::Load,
     },
 };
 
 #[cfg(feature = "dump_load_files")]
-use crate::utils::load::LoadFromFragments;
+use std::path::PathBuf;
```
ClausHolbechArista: these changes are unrelated. See pr 88 and approve that :)

rust/avdschema/src/resolve/resolve_ref.rs
```diff
@@ -21,7 +21,7 @@ pub fn resolve_ref<'a>(ref_: &str, store: &'a Store) -> Result<&'a AnySchema, Sc
     let syntax_err = || RefSyntax {
         schema_ref: ref_.to_owned(),
     };
-    let captures = REF_REGEX.captures(ref_).ok_or_else(syntax_err)?;
+    let captures = REF_REGEX.captures(ref_).ok().flatten().ok_or_else(syntax_err)?;
```
ClausHolbechArista: this must be possible to simplify instead of this long chain.

ClausHolbechArista: so we want to return the new types of errors that might come from captures(). You may need to add or update a regex variant on the SchemaResolverError.

rust/avdschema/src/schema/any.rs
```diff
@@ -12,12 +10,14 @@ use crate::{
...
+#[cfg(feature = "dump_load_files")]
+use crate::utils::load::{LoadError, LoadFromFragments};
```
Shivani-gslab: this change is to avoid the warning for imported but not used

rust/avdschema/src/schema/any.rs
Shivani-gslab: Yes, I came across those warnings while testing.
Will revert these changes. Aprroved the PR #88

rust/avdschema/src/resolve/resolve_ref.rs
Shivani-gslab: Updated the code

rust/avdschema/src/resolve/errors.rs
```diff
@@ -13,6 +13,7 @@ use super::walker::SchemaWalkError;
 pub enum SchemaResolverError {
     SchemaType(SchemaType),
     RefSyntax(RefSyntax),
+    RefRegexError(RefRegexError),
```
ClausHolbechArista:
```suggestion
    RefRegex(RefRegexError),
```

rust/avdschema/src/resolve/resolve_ref.rs
```diff
@@ -126,4 +129,20 @@ mod tests {
...
+    // without injecting a different regex. We verify the error type, display format, and
+    // From conversion here instead.
+    fn ref_regex_error_display() {
+        let err = RefRegexError::new("some_schema#/keys/key".to_owned(), "BacktrackLimitExceeded".to_owned());
+        assert_eq!(
+            err.to_string(),
+            "Regex error for schema $ref 'some_schema#/keys/key': BacktrackLimitExceeded."
+        );
+        let resolver_err: SchemaResolverError = err.into();
+        assert!(matches!(resolver_err, SchemaResolverError::RefRegexError(_)));
```
ClausHolbechArista:
```suggestion
        assert!(matches!(resolver_err, SchemaResolverError::RefRegex(_)));
```

rust/avdschema/src/resolve/errors.rs
ClausHolbechArista: The variants should not repeat error since the enum itself is called error. (Yes I know we have two others that should be renamed :)

Shivani-gslab: should I rename them as well? in this PR

ClausHolbechArista: no

rust/avdschema/src/schema/str.rs
```diff
@@ -4,7 +4,7 @@
 
 use std::sync::OnceLock;
 
-use regex::Regex;
+use fancy_regex::Regex;
```
ClausHolbechArista, line 7: hmm maybe we should revert the changes on the schema ref side and only keep them for the str patterns. Since refs will never need lookaround

Shivani-gslab: as discussed, using fancy-regex to keep single dependency for regexes. it will analyse and  use the NFA engine for simple regex patterns
Reverted the RefRegex error  and the captures part, as it will never fail for refs pattern

rust/validation/src/validation/str.rs
```diff
@@ -72,7 +72,7 @@ fn validate_max_length(schema: &Str, input: &str, ctx: &mut Context) {
 fn validate_pattern(schema: &Str, input: &str, ctx: &mut Context) {
     if let Some(pattern) = &schema.pattern {
         let regex_pattern = pattern.get_compiled_pattern();
-        if !regex_pattern.is_match(input) {
+        if !regex_pattern.is_match(input).unwrap_or(false) {
```
gmuloc: I am not sure we want to merge together Regex error and NotMatchingPattern as it can be very confusing for the user.
I would rather we handle the Result from is_match properly

gmuloc (changes requested):

rust/validation/src/validation/str.rs
Shivani-gslab: To handle them separately, we can match on the `Result` from `is_match` — mapping `Ok(false)` to `Violation::NotMatchingPattern `and `Err(e) to ErrorIssue::InternalError { message: e.to_string() }`.
or a new `Violation::RegexError` variant

Let me know if that works.

ClausHolbechArista: An error on the regex here should be considered an InternalError since the regex comes from the schema. If we already have an error class for invalid schema data we can use that. If not maybe we should add one like ErrorIssue::SchemaRegex(The regex crate error enum here)

Shivani-gslab: we have ` ErrorIssue::InternalError ` in feedback.rs

rust/validation/src/validation/str.rs
```diff
@@ -355,4 +357,163 @@ mod tests {
             }]
         );
     }
+
+    // --- lookaround tests (require fancy-regex) ---
+
+    #[test]
+    fn validate_pattern_positive_lookahead_ok() {
```
gmuloc: are you not essentillay testing fancy-regex lookahead? I mean what's the purpose of these tests for our own crate?
Maybe you can keep only one to check that failed lookahead do provide the violation but doing it for everysingle cases seem overreaching as you are only testing again and again if a non match produce a violation right?

Shivani-gslab: You're right, those tests are redundant.
Let me update the tests.

gmuloc (approved):

ClausHolbechArista (approved):

PR #90, opened 2026-04-14
Fix(rust): Fixing the compilation errors and warnings in avdschema

Shivani-gslab:
## Related Issue(s)

Fixes #<ISSUE ID>

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/rust/avdschema/src/utils/test_utils.rs
+++ b/rust/avdschema/src/utils/test_utils.rs
@@ -4,10 +4,10 @@
 
 use crate::{Store, any::AnySchema};
 
-#[cfg(feature = "dump_load_files")]
-use std::sync::OnceLock;
 use serde::Deserialize as _;
 use serde_json::json;
+#[cfg(feature = "dump_load_files")]
+use std::sync::OnceLock;
 
 // Using a tmp path in the crate allows us to inspect the generated artifacts.
 // The files in the path are exempted from git.
@@ -260,10 +260,7 @@ static AVD_STORE: OnceLock<Store> = OnceLock::new();
 fn init_avd_store() -> Store {
     use crate::Load as _;
     let bytes = std::fs::read(test_schema_store::get_store_gz_path()).unwrap();
-    Store::from_gz_bytes(&bytes)
-        .unwrap()
-        .as_resolved()
-        .unwrap()
+    Store::from_gz_bytes(&bytes).unwrap().as_resolved().unwrap()
 }
 
 #[cfg(feature = "dump_load_files")]
--- a/rust/avdschema/src/utils/test_utils.rs
+++ b/rust/avdschema/src/utils/test_utils.rs
@@ -2,9 +2,10 @@
 // Use of this source code is governed by the Apache License 2.0
 // that can be found in the LICENSE file.
 
-use std::sync::OnceLock;
+use crate::{Store, any::AnySchema};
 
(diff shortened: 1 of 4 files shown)


rust/avdschema/src/utils/test_utils.rs
```diff
@@ -247,16 +252,22 @@ pub(crate) fn get_test_dict_schema() -> AnySchema {
     .unwrap()
 }
 
+// AVD_STORE, init_avd_store and get_avd_store are always compiled (not gated behind
+// dump_load_files) to satisfy #![deny(unused_crate_dependencies)] for test_schema_store.
+// They are only called from dump_load_files-gated tests, hence the allow(dead_code).
+#[allow(dead_code)]
```
ClausHolbechArista: Better to just move these under the feature gate. But it looks like this whole file more or less is gated by that feature? If so, just gate the `mod test_utils` in the `mod.rs`

ClausHolbechArista: I think you can make the test_schema_store depend on the feature as well somehow.

Shivani-gslab: Updated!

rust/avdschema/src/utils/test_utils.rs
```diff
@@ -2,19 +2,25 @@
...
-use crate::{Load as _, Store, any::AnySchema};
 use serde::Deserialize as _;
 use serde_json::json;
+#[cfg(feature = "dump_load_files")]
+use std::sync::OnceLock;
 
 // Using a tmp path in the crate allows us to inspect the generated artifacts.
 // The files in the path are exempted from git.
+#[cfg(feature = "dump_load_files")]
+use std::path::PathBuf;
```
gmuloc: why are we not using a single statement

```
#[cfg(feature = "dump_load_files")]
use std::{path::PathBuf, sync::OnceLock};
```

rust/avdschema/src/utils/test_utils.rs
```diff
@@ -2,19 +2,25 @@
...
+#[cfg(feature = "dump_load_files")]
 const TMP_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tmp");
 
+#[cfg(feature = "dump_load_files")]
```
gmuloc, line 16: it looks like rather than sprinkling things around you could use this pattern:

```
// container
#[cfg(feature = "dump_load_files")]
mod dump_utils {
    pub fn get_tmp_path() -> PathBuf { ... }
    // ... other stuff
}

// Re-export
#[cfg(feature = "dump_load_files")]
pub(crate) use dump_utils::get_tmp_path;
```

gmuloc: @ClausHolbechArista for visibility

ClausHolbechArista: if we go there please do it as a separate file instead of an inline module, but really I don't think we should spend energy on that. I'd rather move the dump/load trait to a separate crate and keep the implementations there. (later)

ClausHolbechArista (approved):

rust/avdschema/src/utils/test_utils.rs
Shivani-gslab: Combined them, thanks!

gmuloc (approved):

PR #94, opened 2026-04-22
Refactor(validation): abstract ParseDiagnostic conversions

ClausHolbechArista:
## Change Summary

Abstract ParseDiagnostic conversions and implement for serde_json::Value

## Proposed changes

As preparation for yaml-parser support in validation we need to support different kinds of source locations. In yaml-parser we track bytes. In serde_json they track lines and columns. Yaml-parser provides a source-map to convert from bytes to lines and columns. This will be implemented later as part of the implementation of this trait for yaml-parser.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/rust/validation/src/feedback.rs
+++ b/rust/validation/src/feedback.rs
@@ -185,24 +185,106 @@ pub struct SourceSpan {
     pub end: usize,
 }
 
+/// 1-based line and column location within the original source input.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct LineColumn {
+    pub line: usize,
+    pub column: usize,
+}
+
+/// Parser-native location for an input diagnostic.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub enum DiagnosticLocation {
+    SourceSpan(SourceSpan),
+    LineColumn(LineColumn),
+}
+
+impl LineColumn {
+    fn to_source_span(&self, input: &str) -> SourceSpan {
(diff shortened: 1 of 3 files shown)


rust/validation/src/feedback.rs
```diff
@@ -185,24 +185,106 @@ pub struct SourceSpan {
...
+
+/// Parser-native location for an input diagnostic.
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub enum DiagnosticLocation {
+    SourceSpan(SourceSpan),
+    LineColumn(LineColumn),
+}
+
+impl LineColumn {
+    fn to_source_span(&self, input: &str) -> SourceSpan {
```
gmuloc, line 203: small optimization to get the line starts

```
let line_starts: Vec<usize> = std::iter::once(0)
    .chain(input.match_indices('\n').map(|(i, _)| i + 1))
    .collect();
```

start is then `line_starts[line - 1] + column - 1`

gmuloc: though it is relevant per input (to be cached) so not sure this needs to go in there

gmuloc (approved):
overall this LGTM except for the potential optimisation we can keep for later when there isa good place for this (maybe we track it somewhere)

rust/validation/src/feedback.rs
ClausHolbechArista: created issue #104

PR #96, opened 2026-04-27
Refactor: Handling compilation panicking for regex patterns in avdschema

Shivani-gslab:
## Change Summary

**Problem:**
When a schema pattern field contained a look-around assertion (e.g. a negative lookahead like (?!radius$|tacacs\+$|ldap$).+), validation would panic at runtime instead of reporting an error:

**thread panicked at rust/avdschema/src/schema/str.rs:105:81:
called `Result::unwrap()` on an `Err` value: Syntax(
    regex parse error:
        ^^(?!radius$|tacacs\+$|ldap$).+$$
    error: look-around, including look-ahead and look-behind, is not supported
)**
Pattern::get_compiled_pattern() called .unwrap() on Regex::new(...), so any compilation failure — including patterns that fancy-regex rejects — caused a hard panic in the validation thread.

**Proposed Solution:**
**rust/avdschema/src/schema/str.rs** — Changed compiled_pattern from OnceLock<Regex> to OnceLock<Result<Regex, String>>. get_compiled_pattern() now returns Result<&Regex, &str> instead of &Regex, propagating compilation errors instead of panicking.

**rust/validation/src/validation/str.rs** — Updated validate_pattern to handle the Result from get_compiled_pattern(). Compilation errors are reported as InternalError (same path already used for is_match runtime errors), so validation degrades gracefully rather than crashing.
## Related Issue(s)

Fixes #<ISSUE ID>

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/rust/avdschema/src/schema/str.rs
+++ b/rust/avdschema/src/schema/str.rs
@@ -85,13 +85,18 @@ impl<'x> TryFrom<&'x AnySchema> for &'x Str {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, derive_more::Display)]
+#[derive(Debug, Serialize, Deserialize, derive_more::Display)]
 #[display("{pattern}")]
 #[serde(transparent)]
 pub struct Pattern {
     pub pattern: String,
     #[serde(skip)]
-    compiled_pattern: OnceLock<Result<Regex, String>>,
+    compiled_pattern: OnceLock<Result<Regex, fancy_regex::Error>>,
+}
+impl Clone for Pattern {
+    fn clone(&self) -> Self {
+        Self::new(self.pattern.clone())
+    }
 }
 impl Pattern {
     fn new(pattern: String) -> Self {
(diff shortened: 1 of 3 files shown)


rust/validation/src/validation/str.rs
```diff
@@ -116,22 +116,16 @@ fn validate_max_length<V: ValidatableValue>(
...
 fn validate_pattern<V: ValidatableValue>(schema: &Str, value: &V, input: &str, ctx: &mut Context) {
     if let Some(pattern) = &schema.pattern {
-        let regex_pattern = pattern.get_compiled_pattern();
-        match regex_pattern.is_match(input) {
-            Ok(true) => {}
-            Ok(false) => ctx.add_error_for(
-                value,
-                Violation::NotMatchingPattern {
+        match pattern.get_compiled_pattern() {
+            Err(e) => ctx.add_error_for(value, ErrorIssue::InternalError { message: e.to_string() }),
```
ClausHolbechArista: these are validation errors, but the compilation error has nothing to do with the user input. It is something from our schema not compiling properly. Let's make the error message clear at least.

rust/avdschema/src/schema/str.rs
```diff
@@ -100,9 +100,11 @@ impl Pattern {
             compiled_pattern: Default::default(),
         }
     }
-    pub fn get_compiled_pattern(&self) -> &Regex {
+    pub fn get_compiled_pattern(&self) -> Result<&Regex, &str> {
         self.compiled_pattern
-            .get_or_init(|| Regex::new(format!("^{}$", &self.pattern).as_str()).unwrap())
+            .get_or_init(|| Regex::new(format!("^{}$", &self.pattern).as_str()).map_err(|e| e.to_string()))
+            .as_ref()
+            .map_err(String::as_str)
```
ClausHolbechArista: why not just change the return err to be the error type returned from regex. That will make it more clear and we can decide how to convert in the calling code.

rust/validation/src/validation/str.rs
Shivani-gslab: Updated it

rust/avdschema/src/schema/str.rs
Shivani-gslab: updated

ClausHolbechArista (approved):

gmuloc (approved):

PR #98, opened 2026-04-30
Refactor: Consolidate input validation in simple_7 encrypt/decrypt

Shivani-gslab:
## Change Summary
Resolves a TODO - https://github.com/aristanetworks/avd/pull/6903
to consolidate input validation for simple_7 functions at the
Python boundary, rather than relying on individual callers (tacacs, ntp, etc.)
to perform their own checks.

## Related Issue(s)

Fixes #<ISSUE ID>

### User Checklist

- N/A

### Repository Checklist


rust/pypasswords/src/tests/test_simple_7.rs
```diff
@@ -59,11 +53,10 @@ fn simple_7_encrypt_invalid_salt_err() {
             .call_method1("simple_7_encrypt", (password, Some(invalid_salt)))
             .unwrap_err();
 
-        // Maps Simple7Error::InvalidSaltValue -> PyValueError
         assert!(err.is_instance_of::<pyo3::exceptions::PyValueError>(py));
         assert_eq!(
             err.value(py).to_string(),
-            "Salt must be in the range 0-15, got 16"
+            "Salt MUST be an integer within the range 0-15."
```
gmuloc: don't use `MUST` thats a remnant of me being RFC-brainwashed

gmuloc: the previous error message was better because it also told you what was there.

This is ok to update the test in pyavd to catch this new error instead

rust/pypasswords/src/lib.rs
```diff
@@ -80,8 +95,17 @@ mod passwords {
     #[cfg(feature = "simple-7")]
     #[pyfunction]
     /// Decrypt (deobfuscate) a password from insecure type-7.
-    pub fn simple_7_decrypt(data: String) -> PyResult<String> {
-        passwords::simple_7_decrypt(&data).map_err(|err| match err {
+    ///
+    /// Raises ValueError if password is not a non-empty string or decryption fails.
+    pub fn simple_7_decrypt(data: Bound<'_, PyAny>) -> PyResult<String> {
+        let data_str: String = data
```
gmuloc: You don't need really the object (and then ajust inputs to simple_7_decrypt
```suggestion
        let data_str: &str = data
```

rust/pypasswords/src/lib.rs
```diff
@@ -69,9 +70,23 @@ mod passwords {
     #[pyfunction]
     /// Encrypt (obfuscate) a password with insecure type-7.
     ///
-    /// If salt is None, a random salt in the range 0-15 will be used.
-    pub fn simple_7_encrypt(data: String, salt: Option<u8>) -> PyResult<String> {
-        passwords::simple_7_encrypt(&data, salt).map_err(|err| match err {
+    /// Raises ValueError if password is not a non-empty string or salt is not an integer in 0-15.
+    pub fn simple_7_encrypt(data: Bound<'_, PyAny>, salt: Bound<'_, PyAny>) -> PyResult<String> {
```
gmuloc: so we don't really want to relax the typing in the library (string) to become PyAny _just_ for the error message as it is a bit sad (relaxing typing to recheck types..).
Let's investigate if we can get a better error message from PyO3 :).
discussed with @ClausHolbechArista who told me he would take a look but feel free to look yourself.

rust/pypasswords/src/lib.rs
```diff
@@ -69,9 +70,23 @@ mod passwords {
...
+        let data_str: String = data
+            .extract()
+            .map_err(|_| PyValueError::new_err("Password MUST be a string with at least 1 character."))?;
+        if data_str.is_empty() {
```
gmuloc: this checks should move to the passwords crate rather than pypassword to avoid the bindings behaving too differently (and then it should bubble up)

rust/pypasswords/src/lib.rs
Shivani-gslab: I have updated the code, Please revisit

rust/pypasswords/src/lib.rs
Shivani-gslab: moved

rust/passwords/src/simple_7/mod.rs
```diff
@@ -18,11 +18,16 @@ pub enum Simple7Error {
...
+    #[display("Password must not be empty")]
+    EmptyPassword,
 }
 impl std::error::Error for Simple7Error {}
 
 /// Decrypt (deobfuscate) a password from insecure type-7.
 pub fn simple_7_decrypt(data: &str) -> Result<String, Simple7Error> {
+    if data.is_empty() {
+        return Err(Simple7Error::EmptyPassword);
+    }
```
ClausHolbechArista: this is not empty password. The data is the encrypted string and we already catch the length below

gmuloc (approved):
Looks ok to me - the coverage thing is a bit suprising as it is showing a line with only `}` as not covered. Wondering why this is.

rust/passwords/src/simple_7/mod.rs
Shivani-gslab: Yeah my bad, its decrypt and datatooshort is already there.
removed it

PR #100, opened 2026-05-04
CI: Lock toolchain and enforce import grouping and sorting

ClausHolbechArista:
## Change Summary

- Lock toolchain to ensure consistent versioning
- Use nightly toolchain for rustfmt to support grouping of imports
- Ran pre-commit and some cleanup to address clippy not liking {self} imports.

## Component(s) name

rust

### User Checklist

- N/A

### Repository Checklist


.github/workflows/pull-request-management.yml
```diff
@@ -20,6 +20,8 @@ env:
   PY_COLORS: 1
   # rust
   CARGO_TERM_COLOR: always
+  CARGO_DENY_VERSION: "0.19.4"
+  CARGO_LLVM_COV_VERSION: "0.8.5"
```
gmuloc: is this something we need to maintain here rather than in some Cargo.toml?

.github/workflows/pull-request-management.yml
```diff
@@ -78,18 +80,23 @@ jobs:
           - macos-latest
     steps:
       - uses: actions/checkout@v6
-        # This is needed for arm machines since they don't ship with 2024 edition.
-      - name: Upgrade rust and cargo
-        run: rustup update stable
-        # This is needed because we have pyo3 tests and our minimum python version is 3.10.
-        # On some test containers (looking at you windows-latest) the included Python is 3.9.
+      - name: Install Rust toolchain from rust-toolchain.toml
+        run: rustup show
```
gmuloc, line 82: is show really doing an installation?

.pre-commit-config.yaml
```diff
@@ -60,12 +60,23 @@ repos:
...
+  - repo: local
+    hooks:
+      - id: cargo-fmt
+        name: Lint Rust files using fmt (nightly)
+        entry: env RUSTUP_TOOLCHAIN=nightly cargo fmt
+        args:
+          - --all
+          - --
+          - --config
+          - unstable_features=true,group_imports=StdExternalCrate,imports_granularity=Item
```
gmuloc, line 74: why not do this in the rustfmt.toml?

.github/workflows/pull-request-management.yml
```diff
@@ -78,18 +80,23 @@ jobs:
...
-      - name: Check code formatting with cargo fmt
-        run: cargo fmt --verbose
+      - name: Check code formatting with cargo fmt (nightly)
+        run: >-
+          RUSTUP_TOOLCHAIN=nightly cargo fmt
+          --verbose
+          --check
+          --all
+          --
+          --config unstable_features=true,group_imports=StdExternalCrate,imports_granularity=Item
```
gmuloc: not sure why we don't do this from rustfmt.yml. That seems like many places to remember to update

.github/workflows/pull-request-management.yml
gmuloc: ok after reading I udnerstand it wil read rust-toolchain.toml and update so thats why.

.github/dependabot.yml
```diff
@@ -26,3 +26,19 @@ updates:
       day: "monday"
     commit-message:
       prefix: "CI: "
+  # Maintain dependencies for Rust crates
+  - package-ecosystem: "cargo"
```
gmuloc, line 30: no grouping on this one? (open question :) ) - I think probably not as we want to control what we change

.github/workflows/pull-request-management.yml
gmuloc: I read we could put them in dev-dependencies of the worksapce and it would be updated by dependabot

rust-toolchain.toml
```diff
@@ -0,0 +1,7 @@
+[toolchain]
+channel = "1.95.0"
```
gmuloc, line 2: hmmm so if I understand correctly we are having this here but then running nighlty everywhere? I am not too sure about all this. Maybe we can discuss?

.github/dependabot.yml
ClausHolbechArista: I think it is fine to keep separate. TBH the groups annoy me in general :)

rust-toolchain.toml
ClausHolbechArista: as discussed but adding here for reference, in pre-commit and CI we are running nightly but on dev laptops and IDEs we may be running stable, so we cannot rely on the nightly-only features in the general fmt settings.

.pre-commit-config.yaml
ClausHolbechArista: as discussed but adding here for reference, in pre-commit and CI we are running nightly but on dev laptops and IDEs we may be running stable, so we cannot rely on the nightly-only features in the general fmt settings.

.github/workflows/pull-request-management.yml
ClausHolbechArista: as discussed but adding here for reference, in pre-commit and CI we are running nightly but on dev laptops and IDEs we may be running stable, so we cannot rely on the nightly-only features in the general fmt settings.

.github/dependabot.yml
gmuloc: ack

.github/workflows/pull-request-management.yml
ClausHolbechArista: moving cargo-deny to pre-commit.
Removing the version of llvm-cov so we just use the latest.

Cargo.lock
gmuloc, line 1: you can remove the .sonarcloud.config file as we are nice citizens now

```diff
- # Ignore text rule S8570
- sonar.issue.ignore.multicriteria=ignore_s8570
- sonar.issue.ignore.multicriteria.ignore_s8570.ruleKey=text:S8570
- sonar.issue.ignore.multicriteria.ignore_s8570.resourceKey=**/Cargo.toml
```

gmuloc (approved):

PR #105, opened 2026-05-04
Refactor(rust): Implement schemadatavalue for yaml parser value

ClausHolbechArista:
Add implementation of avdschema::SchemaDataValue (and associated views) for the yaml_parser::Value.

This will *not* promote the yaml_parser to be actively used.
It is just an implementation to be used later.

### DIFF
--- a/rust/yaml-parser/src/avdschema_data_impl.rs
+++ b/rust/yaml-parser/src/avdschema_data_impl.rs
@@ -1,0 +1,397 @@
+// Copyright (c) 2026 Arista Networks, Inc.
+// Use of this source code is governed by the Apache License 2.0
+// that can be found in the LICENSE file.
+
+use avdschema::{SchemaDataMapping, SchemaDataSequence, SchemaDataValue};
+
+use crate::{MappingPair, SequenceItem, Value};
+
+#[derive(Debug, Clone, Copy)]
+pub struct YamlMapping<'a, 'input>(&'a [MappingPair<'input>]);
+
+#[derive(Debug, Clone, Copy)]
+pub struct YamlSequence<'a, 'input>(&'a [SequenceItem<'input>]);
+
+impl<'a, 'input: 'a> SchemaDataValue<'a> for &'a Value<'input> {
+    type Mapping = YamlMapping<'a, 'input>;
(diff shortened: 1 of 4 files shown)


rust/yaml-parser/src/lib.rs
```diff
@@ -186,6 +188,8 @@ pub mod writer;
 pub mod serde;
 
 // Public API: high-level parsing, AST, spans, and errors.
+#[cfg(feature = "avdschema")]
+pub use avdschema_data_impl::YamlMapping;
```
gmuloc, line 192: not blocking: while today all our schema are dicts we could consider also exposing YamlSequence here

rust/yaml-parser/src/avdschema_data_impl.rs
```diff
@@ -0,0 +1,420 @@
...
+#[derive(Debug, Clone, Copy)]
+/// Schema-data mapping view over a YAML mapping's raw pair slice.
+pub struct YamlMapping<'a, 'input>(&'a [MappingPair<'input>]);
+
+#[derive(Debug, Clone, Copy)]
+pub struct YamlSequence<'a, 'input>(&'a [SequenceItem<'input>]);
+
+impl<'a, 'input> YamlMapping<'a, 'input> {
+    /// Create a schema-data mapping view from a YAML mapping pair slice.
+    pub const fn new(pairs: &'a [MappingPair<'input>]) -> Self {
```
gmuloc: I am striggling to understand why we are adding const here

rust/yaml-parser/src/lib.rs
gmuloc: not needed today

ClausHolbechArista: only exposing this one because it is the only one being used (in the validation)

rust/yaml-parser/src/avdschema_data_impl.rs
ClausHolbechArista:
```suggestion
    pub fn new(pairs: &'a [MappingPair<'input>]) -> Self {
```

gmuloc (approved):

PR #106, opened 2026-05-05
Refactor(rust): Improve yaml-parser handling for block scalars

ClausHolbechArista:
## Change Summary

Improve yaml-parser handling for block scalars

## Related Issue(s)

Fixes a bug where this yaml produces indentation and unmatched quotes errors:

```yaml
key: >-
  'foo
```

Also improves performance significantly.

## Component(s) name

rust

## Proposed changes

- Move block scalar handling to the lexer to it can gather the text without stumbling on other token types.
  - It was an issue to keep this in the emitter since lookaheads would trigger tokenization of the content of the block scalar without understanding the context and different rule-set.

## How to test

- Added many tests for various corner cases
- Reran benchmarks and updated doc

### User Checklist

- N/A

### Repository Checklist


rust/yaml-parser/tests/edge_cases.rs
```diff
@@ -805,6 +805,191 @@ fn test_block_scalar_chomping() {
...
+            .as_node()
+            .trailing_comment()
+            .map(|comment| comment.text.as_ref()),
+        Some(" header"),
+    );
+}
+
+#[test]
+fn test_unicode_block_scalar_content_and_header_comment() {
+    let docs = parse_ok("mý🔑: >- # héader ☃\n  café 'quote'\n  emoji 😀\n");
```
gmuloc, line 990: ❤️

rust/yaml-parser/tests/edge_cases.rs
```diff
@@ -805,6 +805,191 @@ fn test_block_scalar_chomping() {
     }
 }
 
+#[test]
+fn test_folded_block_scalar_quotes_are_content() {
```
gmuloc, line 867: not sure if we are adding this here but look slike we are not testing empty lines before first content when we are tracking them

rust/yaml-parser/src/lexer/block_scalars.rs
gmuloc, line 1: suggested extra tests for regression testing



  - Tabs around block scalar indentation
```
  key: |
        foo
```
  and:
```
  key: |
        foo
```
  The helper counts only spaces as indent and treats tabs as content or invalid depending on position. That is intentional-looking, but it is a classic YAML edge case.

  - Indented document-marker-looking content
```
  key: |-
    ...
```
  Only zero-indented ... / --- should terminate the scalar. The helper checks indent == 0, so this should be fine, but it’s a useful regression case.

  - Header comment plus following sibling
```
  key: | # header
    value
  next: value
```
  The header comment is queued as a pending token while the block scalar token is returned first. The new comment tests cover attachment, but a sibling-after-scalar case would also check pending token ordering.

  - Explicit indent in nested contexts
```
  outer:
    key: |1
     value
```

all of these are handled correctly by the current PR code.

rust/yaml-parser/src/lexer/block_scalars.rs
```diff
@@ -91,4 +136,335 @@ impl<'input> Lexer<'input> {
...
+                    self.add_error(ErrorKind::InvalidIndentation, line_start_span);
+                }
+                self.pending_tokens.push_back(super::RichToken::new(
+                    Token::LineStart(crate::span::usize_to_indent(indent)),
+                    line_start_span,
+                ));
+                self.byte_pos = line_start;
+                break;
+            }
+            if let Some(ci) = content_indent
```
gmuloc, line 237: this long functions would benefit from comments to explain the decisions taken (here indent to small, exist the block and move on to next line)

gmuloc (commented):
overall ok with the PR - adding a couple of tests suggestion that may be in the YAML suite but would help understand locally what we are doing

rust/yaml-parser/src/lexer/block_scalars.rs
ClausHolbechArista: added

rust/yaml-parser/tests/edge_cases.rs
ClausHolbechArista: added tests, from your comment it looks like you are expecting us to trim the leading empty lines?

rust/yaml-parser/src/lexer/block_scalars.rs
```diff
@@ -91,4 +136,366 @@ impl<'input> Lexer<'input> {
...
+                }
+            }
+
+            let ci = content_indent.unwrap_or(indent);
+            let extra_indent = indent.saturating_sub(ci);
+            let has_non_whitespace = line_text.bytes().any(|byte| !matches!(byte, b' ' | b'\t'));
+            // A tab in the visible prefix is content only after the content
+            // indent; before that it is an invalid indentation character.
+            if has_tab_in_prefix && indent < content_indent.unwrap_or(min_auto_indent) {
+                self.add_error(ErrorKind::InvalidIndentation, line_start_span);
```
gmuloc: this is uncoverted according to codecov and this looks to be unreachable.

rust/yaml-parser/tests/edge_cases.rs
gmuloc:
```
# adding \n for correctness
key: |\n
    \n
  value\n
```

I think this should report invalid indentation but checking codecov ( :) )  I see it is already tested and trqcked it down to `W9L4` from the main suite and it is triggering the InvalidIndentation. So we can resolve this. Thanks for the extra tests

rust/yaml-parser/src/lexer/block_scalars.rs
ClausHolbechArista: removed as indeed it is caught earlier.

gmuloc (approved):

PR #109, opened 2026-05-05
Fix(rust): YAML parser - fix incorrect bridging to dedented value.

ClausHolbechArista:
The YAML

```yaml
mylist:
  - key1: foo
    key2:
  - key1: bar
```
Triggered that key2 got loaded with a list starting the the next item instead of being loaded with `null`.
Fixed for sequence and map.

### DIFF
--- a/rust/yaml-parser/tests/edge_cases.rs
+++ b/rust/yaml-parser/tests/edge_cases.rs
@@ -904,7 +904,7 @@ catalog:
     let docs = parse_ok(input);
     assert_eq!(docs.len(), 1);
 
-    let root_pairs = match &docs.first().expect("expected exactly one document").value {
+    let root_pairs = match &docs[0].value {
         Value::Mapping(pairs) => pairs,
         other => panic!("expected root mapping, got {other:?}"),
     };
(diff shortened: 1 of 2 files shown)


rust/yaml-parser/tests/edge_cases.rs
```diff
@@ -888,6 +888,125 @@ fn test_mixed_flow_and_block() {
...
+    let catalog_pairs = match &catalog.value.value {
+        Value::Mapping(pairs) => pairs,
+        other => panic!("expected catalog mapping, got {other:?}"),
+    };
+    let items = catalog_pairs
+        .iter()
+        .find(|pair| pair.key.value == Value::String("items".into()))
+        .expect("expected items key");
+    let item_nodes = match &items.value.value {
+        Value::Sequence(items) => items,
```
gmuloc: this is introducing a clippy warning for shadow reuse so maybe owrth changing the name?

gmuloc (approved):
LGTM

rust/yaml-parser/tests/edge_cases.rs
```diff
@@ -888,6 +888,125 @@ fn test_mixed_flow_and_block() {
...
+        other => panic!("expected items sequence, got {other:?}"),
+    };
+
+    assert_eq!(
+        item_nodes.len(),
+        2,
+        "the second entry must remain a sibling item"
+    );
+
+    let first_item_pairs = match &item_nodes.first().expect("expected first item").node.value {
```
gmuloc: given that we asserte length is 2, we don't really need the first / expect and this would clean the test files a bit

PR #111, opened 2026-05-06
Refactor: Add specific Python Error in pyo3 crates

gmuloc:
## Change Summary

Move Rust-backed Python wrappers to typed, module-specific exception hierarchies.

## Component(s) name

`rust`, `pyavd-utils`

## Proposed changes

- Added Python-specific exception hierarchies for `pyavd_utils.passwords`:
  - Base: `PasswordError`
  - Variant-specific children for SHA512 crypt, CBC, and Simple7 failures.
- Added Python-specific exception hierarchy for `pyavd_utils.validation`:
  - Base: `ValidationError`
  - Store init/load errors, schema resolve/store errors, JSON errors, and internal validation errors.
- Mapped exposed Rust errors through `ToPythonError` implementations in the PyO3 crates.
- Updated `.pyi` stubs to expose and document the new exception classes.
- Set PyO3 exception module paths to the public import paths so classes display as `pyavd_utils.passwords.*` / `pyavd_utils.validation.*` and can be pickled.

## Compatibility note

The new typed exceptions intentionally replace the previous broad `ValueError` / `RuntimeError` behavior. Callers should catch the specific subclasses or the module base classes (`PasswordError` / `ValidationError`) instead of relying on the old built-in exception classes.

## How to test

- Updated Rust and Python tests to assert specific exception types, base-class hierarchy, public `__module__` paths, and pickling behavior.
- Split pyvalidation Rust test setup into Python-only setup and schema-store setup where appropriate.

## Checklist

- [x] I have read the CONTRIBUTING document.
- [x] I have updated testing accordingly.

## Release Notes

* **New Features**
  * Added structured, dedicated Python exception hierarchies for password operations (SHA-512 crypt, CBC, Simple-7) and schema validation, replacing generic errors with specific failure types.

* **Bug Fixes**
  * Improved exception translation to consistently raise the correct specific classes, with updated message wording/punctuation for clearer reporting.

* **Tests**
  * Strengthened coverage to assert exception inheritance, exact exception types/messages, public module paths, and successful pickling/unpickling of the raised exceptions.


pyavd_utils/passwords.pyi
```diff
@@ -4,6 +4,28 @@
 # For now we allow docstrings in stubs
 # ruff: noqa: PYI021
 
+class PyAVDUtilsPasswordError(Exception): ...
+class PyAVDUtilsSha512CryptError(PyAVDUtilsPasswordError): ...
+class PyAVDUtilsSha512CryptInvalidSaltError(PyAVDUtilsSha512CryptError): ...
+class PyAVDUtilsSha512CryptInvalidSaltEmptyError(PyAVDUtilsSha512CryptInvalidSaltError): ...
+class PyAVDUtilsSha512CryptInvalidSaltCharacterError(PyAVDUtilsSha512CryptInvalidSaltError): ...
```
ClausHolbechArista: I don't think we need to prefix every error with the PyAVDUtils origin.

rust/pypasswords/src/lib.rs
```diff
@@ -10,52 +10,243 @@ use pyo3::pymodule;
...
-        exceptions::{PyRuntimeError, PyValueError},
+        Bound, PyResult, create_exception,
+        exceptions::PyException,
         pyfunction,
+        types::{PyModule, PyModuleMethods},
     };
 
+    pub(crate) trait ToPythonError {
+        fn to_python_error(self) -> pyo3::PyErr;
+    }
```
ClausHolbechArista: if you use `impl From<our_type> for PyErr` trait, the compiler will accept ? in the code since it can be converted to PyErr automatically. That will simplify things elsewhere

rust/pypasswords/src/lib.rs
```diff
@@ -10,52 +10,243 @@ use pyo3::pymodule;
...
+
+    macro_rules! define_python_exceptions {
```
ClausHolbechArista: If you the create_exception! outside the module this extra registration should not be needed. So we can just use the create_exception! alone.

rust/pypasswords/src/lib.rs
gmuloc:
```
  error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
     --> rust/pypasswords/src/lib.rs:146:5
      |
  146 |     impl From<passwords::Sha512CryptError> for pyo3::PyErr {
      |     ^^^^^---------------------------------^^^^^-----------
      |          |                                     |
      |          |                                     `PyErr` is not defined in the current crate
      |          `Sha512CryptError` is not defined in the current crate
      |
      = note: impl doesn't have any local type before any uncovered type parameters
      = note: for more information see https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules
      = note: define and implement a trait or new type instead
```

so either I use this trait or I need to make a type - what would you prefer?

rust/pypasswords/src/lib.rs
gmuloc: I will check but I still need to do `m.add()` in the module I believe, hence why I ended up with this macro that both create and add to the module.

gmuloc: adjusted with some annotation. Let me know what you think of this style

rust/pyvalidation/src/lib.rs
```diff
@@ -11,26 +11,250 @@
...
-        Bound, PyResult, exceptions::PyRuntimeError, pyclass, pyfunction, pymethods,
-        types::PyModule,
-    };
+    use pyo3::{Bound, PyResult, pyclass, pyfunction, pymethods, types::PyModule};
     use std::path::PathBuf;
     use validation::{
         Context, StoreValidateInput as _, Validation as _, feedback::InputDiagnostic,
     };
 
+    #[pymodule_export]
```
gmuloc: try using a brace

rust/pyvalidation/src/lib.rs
```diff
@@ -11,26 +11,250 @@
 use std::sync::OnceLock;
 
 use avdschema::Store;
-use pyo3::pymodule;
+use pyo3::{create_exception, exceptions::PyException, pymodule};
 
 static STORE: OnceLock<Store> = OnceLock::new();
 
+create_exception!(
```
gmuloc: move this to a file for exceptions

rust/pypasswords/src/lib.rs
```diff
@@ -3,59 +3,274 @@
 // that can be found in the LICENSE file.
 #![deny(unused_crate_dependencies)]
 
-use pyo3::pymodule;
+use pyo3::{create_exception, exceptions::PyException, pymodule};
+
+create_exception!(
+    passwords,
+    PyAVDUtilsPasswordError,
```
ClausHolbechArista:
```suggestion
    PasswordError,
```

rust/pypasswords/src/lib.rs
```diff
@@ -3,59 +3,274 @@
...
+    PyException,
+    "Base exception for pyavd_utils.passwords."
+);
+create_exception!(
+    passwords,
+    Sha512CryptError,
```
ClausHolbechArista: will this ever be raised or is it always one of the underlying variants?

ClausHolbechArista: I think you did this to match the enum level so we can catch the category of errors, but I am not sure we need that?

pyavd_utils/passwords.pyi
```diff
@@ -81,7 +104,8 @@ def simple_7_encrypt(data: str, salt: int | None) -> str:
         str: The encrypted password in type-7 format.
 
     Raises:
-        ValueError: If the salt is not in the range 0-15.
+        Simple7InvalidSaltValueError: If the salt is not in the range 0-15.
+        Simple7EmptyPasswordError: If the password is empty.
```
gmuloc, line 110:
```suggestion
        Simple7InvalidSaltValueError: If the salt is not in the range 0-15.
        Simple7EmptyPasswordError: If the password is empty.
        Simple7RandomSourceUnavailableError: If random salt generation fails.
```

pyavd_utils/passwords.pyi
```diff
@@ -81,7 +104,8 @@ def simple_7_encrypt(data: str, salt: int | None) -> str:
...
     """
```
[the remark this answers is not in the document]
gmuloc: resolved

rust/pypasswords/src/lib.rs
gmuloc: ok I now create locla enums to be able to do this. let me know

rust/pypasswords/src/lib.rs
```diff
@@ -7,57 +7,68 @@
...
-                passwords::Sha512CryptError::InvalidSalt(_)
-                | passwords::Sha512CryptError::Base64InvalidLength(_) => {
-                    PyValueError::new_err(err.to_string())
-                }
-                passwords::Sha512CryptError::ShaCrypt(_) => {
-                    PyRuntimeError::new_err(err.to_string())
-                }
-            }
-        })
+        Ok(sha512_crypt_impl(password, salt)?)
```
gmuloc: use `Into::into`
```suggestion
        Ok(passwords::sha512_crypt(password, salt).map_err(Into::into)?)
```

rust/pypasswords/src/errors.rs
```diff
@@ -0,0 +1,153 @@
...
+use pyo3::PyErr;
+
+use crate::exceptions;
+
+#[cfg(feature = "sha512")]
+#[derive(Debug, derive_more::From)]
+pub(crate) enum Sha512CryptPyError {
+    Sha512Crypt(passwords::Sha512CryptError),
+}
+
```
gmuloc, line 12:
```
#[cfg(feature = "sha512")]
#[derive(Debug, derive_more::From)]
struct Sha512CryptPyError(passwords::Sha512CryptError);
```

gmuloc: do this for single types

ClausHolbechArista: This can also be done for the enums

rust/pypasswords/src/lib.rs
ClausHolbechArista: changed result type instead which simplifies this in https://github.com/aristanetworks/pyavd-utils/pull/111/commits/2664b90268d33c5419ac13a663372600cfe1e605

rust/pypasswords/src/lib.rs
```diff
@@ -53,28 +53,29 @@ mod passwords {
     #[cfg(feature = "sha512")]
     #[pyfunction]
     /// Computes the SHA512 crypt value for the password given the salt
-    pub(crate) fn sha512_crypt(password: &str, salt: &str) -> PyResult<String> {
-        Ok(passwords::sha512_crypt(password, salt).map_err(Into::<Sha512CryptPyError>::into)?)
+    pub(crate) fn sha512_crypt(password: &str, salt: &str) -> Result<String, Sha512CryptPyError> {
```
gmuloc, line 51: oh so it can take a result and it will make it into a PyResult. neat - did not think about that. is this documented somewhere?

rust/pypasswords/src/errors.rs
```diff
@@ -19,65 +19,44 @@ impl From<Sha512CryptPyError> for PyErr {
...
-    fn from(err: CbcEncryptPyError) -> Self {
-        match err {
-            CbcEncryptPyError::Cbc(err) => cbc_error_to_pyerr(&err),
-            CbcEncryptPyError::InvalidBase64Utf8(err) => {
-                exceptions::CBCInvalidBase64Utf8Error::new_err(format!(
-                    "Base64 output contained invalid UTF-8: {err}"
-                ))
-            }
-        }
+impl From<passwords::CbcError> for CbcPyError {
```
ClausHolbechArista, line 30: why this instead of the derive_more?

rust/pypasswords/src/lib.rs
```diff
@@ -60,22 +55,19 @@ mod passwords {
     #[cfg(feature = "cbc")]
     #[pyfunction]
     /// Encrypt the data with CBC `TripleDES`
-    pub(crate) fn cbc_encrypt(password: &str, data: &str) -> Result<String, CbcEncryptPyError> {
+    pub(crate) fn cbc_encrypt(password: &str, data: &str) -> Result<String, CbcPyError> {
         let result_bytes = passwords::cbc_encrypt(password.as_bytes(), data.as_bytes())?;
-        Ok(String::from_utf8(result_bytes)?)
+        Ok(String::from_utf8(result_bytes).map_err(CbcPyError::InvalidBase64Utf8)?)
```
ClausHolbechArista, line 60: This should not be required if you use the derive_more::From. Then a ? should be enough.

PR #124, opened 2026-05-18
Feat(validation): Implement ValidatableValue for yaml_parser values

ClausHolbechArista:
- Implements the ValidatableValue trait for yaml_parser::Value and associated types.
- Adds validate_yaml() to the StoreValidateInput trait.

This is not exposed to Python so it will not be used for pyavd-utils library (yet).

### DIFF
--- a/rust/validation/src/validatable/mod.rs
+++ b/rust/validation/src/validatable/mod.rs
@@ -285,6 +285,7 @@ pub trait ValidatableSequence<'a> {
 // === Implementations ===
 
 mod serde_json_impl;
+mod yaml_parser_impl;
 
 #[cfg(test)]
 mod tests;
--- a/rust/validation/src/validatable/mod.rs
+++ b/rust/validation/src/validatable/mod.rs
@@ -43,6 +43,12 @@ pub trait ValidatableValue: Sized {
     /// For `serde_json::Value`, this is `Value` (same type).
     type Coerced;
 
+    /// The output entry type used when rebuilding a coerced mapping.
+    ///
+    /// JSON can use a simple `(String, Value)` pair, while YAML uses its
+    /// native mapping-pair type so key and pair spans can be preserved.
+    type CoercedMappingItem;
+
     // === Strict type checking (no coercion) ===
 
     /// Returns `true` if this value is null/None.
@@ -194,7 +200,7 @@ pub trait ValidatableValue: Sized {
     fn coerce_sequence(&self, items: Vec<Self::Coerced>) -> Self::Coerced;
 
     /// Create a coerced mapping from coerced key-value pairs.
-    fn coerce_mapping(&self, items: Vec<(String, Self::Coerced)>) -> Self::Coerced;
+    fn coerce_mapping(&self, items: Vec<Self::CoercedMappingItem>) -> Self::Coerced;
 
     /// Clone the value as-is without type coercion.
     /// Used when there's no schema to guide coercion.
@@ -249,13 +255,38 @@ pub trait ValidatableMappingPair<'a> {
     /// The type of values in this mapping pair.
     type Value: ValidatableValue + 'a;
 
-    /// Get the key as a string.
-    fn key(&self) -> Cow<'a, str>;
+    /// Get the key used for schema lookups.
+    ///
+    /// This returns `None` for keys that must not participate in schema
+    /// matching. For example, YAML supports non-string mapping keys, but AVD
+    /// schemas define string keys only, so those YAML keys return `None`.
+    fn schema_key(&self) -> Option<Cow<'a, str>>;
+
+    /// Get a display representation of the key for paths and diagnostics.
+    ///
+    /// This must return a stable, human-readable key string even when
+    /// [`schema_key`](Self::schema_key) returns `None`, so validation can still
+    /// report a useful path for unexpected or ignored keys.
+    fn display_key(&self) -> Cow<'a, str>;
 
     /// Get the mapped value.
     fn value(&self) -> &'a Self::Value;
 
+    /// Build a coerced mapping entry from this pair's original key and a
+    /// coerced value.
+    ///
+    /// This preserves backend-specific key shape for returned coerced data:
+    /// JSON keeps string keys, while YAML can preserve original non-string
+    /// keys and mapping-pair spans.
+    fn coerced_item(
+        &self,
+        value: <Self::Value as ValidatableValue>::Coerced,
+    ) -> <Self::Value as ValidatableValue>::CoercedMappingItem;
+
     /// Get the source span of the key, if available.
+    ///
+    /// Used for diagnostics that should point at the key token itself, such as
+    /// unexpected-key and deprecation feedback.
     fn key_span(&self) -> Option<SourceSpan> {
         None
     }
--- a/rust/validation/src/validatable/mod.rs
+++ b/rust/validation/src/validatable/mod.rs
@@ -25,15 +25,18 @@ use crate::feedback::Type;
 
 // Attempt lossless conversion to i64.
 pub(crate) fn integral_float_to_i64(float: f64) -> Option<i64> {
-    if float.is_finite()
-        && float.fract() == 0.0
-        && float >= i64::MIN as f64
-        && float <= i64::MAX as f64
-    {
-        Some(float as i64)
-    } else {
-        None
+    if !float.is_finite() || float.fract() != 0.0 {
+        return None;
     }
+
+    let value = float as i64;
+    // Positive out-of-range floats saturate to i64::MAX, and `i64::MAX as f64`
+    // rounds up to 2^63, so the round-trip check below would not reject them.
+    if value == i64::MAX && float.is_sign_positive() {
+        return None;
+    }
+
+    (value as f64 == float).then_some(value)
 }
 
 /// A value that can be validated against a schema.
--- a/rust/validation/src/validatable/mod.rs
+++ b/rust/validation/src/validatable/mod.rs
@@ -23,6 +23,19 @@ use avdschema::SchemaDataMapping;
 use crate::feedback::SourceSpan;
 use crate::feedback::Type;
 
+// Attempt lossless conversion to i64.
+pub(crate) fn integral_float_to_i64(float: f64) -> Option<i64> {
+    if float.is_finite()
+        && float.fract() == 0.0
+        && float >= i64::MIN as f64
+        && float <= i64::MAX as f64
+    {
+        Some(float as i64)
+    } else {
+        None
+    }
+}
+
 /// A value that can be validated against a schema.
 ///
 /// This trait abstracts over the common operations needed during validation,
--- a/rust/validation/src/validatable/yaml_parser_impl.rs
+++ b/rust/validation/src/validatable/yaml_parser_impl.rs
@@ -29,6 +29,7 @@ impl<'input> ValidatableValue for Node<'input> {
     where
         Self: 'a;
     type Coerced = Node<'static>;
+    type CoercedMappingItem = MappingPair<'static>;
 
     fn is_null(&self) -> bool {
         matches!(self.value, Value::Null)
@@ -98,7 +99,7 @@ impl<'input> ValidatableValue for Node<'input> {
         match &self.value {
             Value::Mapping(pairs) => {
                 for pair in pairs {
-                    if let Some(key_str) = coerce_key_to_string(&pair.key)
+                    if let Some(key_str) = schema_key_to_string(&pair.key)
                         && key_str == key
                     {
                         return Some(&pair.value);
@@ -155,34 +156,8 @@ impl<'input> ValidatableValue for Node<'input> {
         )
     }
 
-    fn coerce_mapping(&self, items: Vec<(String, Self::Coerced)>) -> Self::Coerced {
-        let pairs: Vec<MappingPair<'static>> = match &self.value {
-            Value::Mapping(original_pairs) => {
-                let mut original_pairs = original_pairs
-                    .iter()
-                    .filter(|pair| is_coercible_key(&pair.key));
-                items
-                    .into_iter()
-                    .map(|(key, value)| {
-                        let (key_span, pair_span) = original_pairs
-                            .next()
-                            .map(|pair| (pair.key.span, pair.pair_span))
-                            .unwrap_or_else(|| fallback_mapping_spans(value.span));
-                        let key_node = Node::new(Value::String(Cow::Owned(key)), key_span);
-                        MappingPair::new(pair_span, key_node, value)
-                    })
-                    .collect()
-            }
-            _ => items
-                .into_iter()
-                .map(|(key, value)| {
-                    let (key_span, pair_span) = fallback_mapping_spans(value.span);
-                    let key_node = Node::new(Value::String(Cow::Owned(key)), key_span);
-                    MappingPair::new(pair_span, key_node, value)
-                })
-                .collect(),
-        };
-        Node::new(Value::Mapping(pairs), self.span)
+    fn coerce_mapping(&self, items: Vec<Self::CoercedMappingItem>) -> Self::Coerced {
+        Node::new(Value::Mapping(items), self.span)
     }
 
     fn clone_to_coerced(&self) -> Self::Coerced {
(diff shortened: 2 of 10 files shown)


rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+                Integer::BigIntStr(s) => Some(s.clone()),
+            },
+            Value::Float(f) => Some(Cow::Owned(f.to_string())),
+            // Using Title case to match Python behavior
+            Value::Bool(b) => Some(Cow::Borrowed(if *b { "True" } else { "False" })),
+            _ => None,
+        }
+    }
+
+    fn as_i64(&self) -> Option<i64> {
```
gmuloc, line 62: as_i64 and is_int can give contradicting information for big integers (e.g. I128, U128 and BigIntStr)

`9223372036854775808` -> Yaml Parser says it is an int but as_i64 says it is None. This may be acceptable but a weird corner case leading to 

```
Expected type int but got int
```

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+            Value::String(_) => Type::Str,
+            Value::Sequence(_) => Type::List,
+            Value::Mapping(_) => Type::Dict,
+        }
+    }
+
+    // === Coercion builders ===
+    // These preserve the span from the original node.
+
+    fn coerce_null(&self) -> Self::Coerced {
```
gmuloc, line 127: here we are dropping comments, anchor, tags,... from the original node. I think it is ok as we are not using anything else today but that could be something to revisit.

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+                let pair_span = yaml_parser::Span::new(key_span.start..value.span.end);
+                let key_node = Node::new(Value::String(Cow::Owned(key)), key_span);
+                MappingPair::new(pair_span, key_node, value)
+            })
+            .collect();
+        Node::new(Value::Mapping(pairs), self.span)
+    }
+
+    fn clone_to_coerced(&self) -> Self::Coerced {
+        self.clone().into_owned()
```
gmuloc, line 160: this would keep the Node metadata so we could end up with mixed flavors of Nodes. (but as I said, this seems ok for now)

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+            self.span,
+        )
+    }
+
+    fn coerce_mapping(&self, items: Vec<(String, Self::Coerced)>) -> Self::Coerced {
+        // Convert string keys to Node keys
+        let pairs: Vec<MappingPair<'static>> = items
+            .into_iter()
+            .map(|(key, value)| {
+                let key_span = yaml_parser::Span::at(value.span.start);
```
gmuloc: feels like we are losing information here

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+                let pair_span = yaml_parser::Span::new(key_span.start..value.span.end);
```
gmuloc: this feels weird to use `key_span.start` which is actually `value.span.start` here as it gives some wrong impression but I guess thats what we are passing today and it looks ok from usage

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -0,0 +1,818 @@
...
+            Value::Sequence(seq) => FV::List(
+                seq.iter()
+                    .map(|item| item.node.to_feedback_value())
+                    .collect(),
+            ),
+            Value::Mapping(map) => FV::Dict(
+                map.iter()
+                    .filter_map(|pair| {
+                        pair.key
+                            .as_str()
```
gmuloc: this could cause some weirdness for bool key as as_str returns "True"/"False" when coerce_key_to_string wuld return "true"/"false"

gmuloc: but we don't use bool keys so we should survive

gmuloc: hmm given the comment below I wonder if that would not be better though to use the same function

gmuloc: showing the inconsistency
```rust
  #[test]
  fn yaml_edge_case_feedback_value_uses_title_case_bool_mapping_key() {
      let mapping = Node::new(
          Value::Mapping(vec![MappingPair::new(
              make_span(),
              Node::new(Value::Bool(true), make_span()),
              string_node("enabled"),
          )]),
          make_span(),
      );

      let feedback_value = mapping.to_feedback_value();

      assert_eq!(
          feedback_value,
          crate::feedback::Value::Dict(std::collections::HashMap::from([(
              "True".to_owned(),
              crate::feedback::Value::Str("enabled".to_owned()),
          )]))
      );

      assert_eq!(
          mapping
              .get("true")
              .and_then(|node| node.as_str())
              .as_deref(),
          Some("enabled")
      );

      assert!(mapping.get("True").is_none());
  }
```

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: Correct, the properties and trailing comment are relatively expensive to clone so unless we need them I think we should avoid that.
I hope we can end up later at a place where we don't have to rebuild all the data for coercion.

ClausHolbechArista: added a comment

rust/validation/src/validation/store.rs
```diff
@@ -31,6 +33,13 @@ pub struct InputValidationResult<T> {
     pub document: ValidationOutput<T>,
 }
 
+#[derive(Debug)]
+/// Result of validating a YAML input source that may contain multiple documents.
+pub struct YamlValidationResult<T> {
```
gmuloc, line 38: would it make sense to have only one InputValidationResult and change the document to documents?

rust/validation/src/validation/store.rs
```diff
@@ -120,9 +171,124 @@ pub enum StoreValidateError {
...
     use super::*;
+    use crate::feedback::Feedback;
     use crate::feedback::ParseDiagnosticKind;
+    use crate::feedback::SourceSpan;
+    use crate::feedback::Type;
+    use crate::feedback::Violation;
     use crate::validation::test_utils::get_test_store;
 
+    #[test]
+    fn validate_yaml_err() {
```
gmuloc, line 182: we could add a test for: empty YAML input, comments-only YAML input and verify the result are what we want.

rust/validation/src/validation/store.rs
```diff
@@ -120,9 +171,124 @@ pub enum StoreValidateError {
...
+                diagnostic,
+                InputDiagnostic::ParseDiagnostic(parse_diagnostic)
+                    if parse_diagnostic.kind == ParseDiagnosticKind::YamlSyntax
+                        && parse_diagnostic
+                            .to_source_span(input)
+                            .end
+                            >= parse_diagnostic.to_source_span(input).start
+            )
+        }));
+        assert!(!output.documents.is_empty());
```
gmuloc, line 238: we don't have a test where we end up with no document. According to my friend this could be an option:

```
let input = "*undefined_alias";
```

rust/validation/src/validation/store.rs
```diff
@@ -120,9 +171,124 @@ pub enum StoreValidateError {
...
+                    expected: Type::Str,
+                    found: Type::Dict
+                }
+                .into()
+            },]
+        );
+    }
+
+    #[test]
+    fn validate_yaml_ok_with_coerced_data() {
```
gmuloc, line 287: could be interesting to make sure we can coeerce multiple documents (we can today but would help catch regressions)

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: what do we want? I chose i64 as the schema type for min/max etc, but I guess we can relax it a bit for type checks and just accept any int type.

rust/validation/src/validation/store.rs
ClausHolbechArista: are those really validation tests or yaml parser tests?

rust/validation/src/validation/store.rs
ClausHolbechArista: well in json there will only be one result and it would break the pyvalidation signature unless we hardcode a documents.first() in there, which also seems bad in a different way.

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: rebuilding this to preserve proper spans

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: we are not using the spans on the coerced data so this was not caught. Rebuilding this to preserve input spans. Closing some of the overlapping comments.

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: hm yeah you are right. The clone is rather expensive compared to just copying the spans, so I can make this consistent with the other to not carry over everything or change the other one to carry over?

rust/validation/src/validation/store.rs
gmuloc: discussed and closing for now as this spearation is stronger.

rust/validation/src/validation/store.rs
gmuloc: we would need to add test for restrict_null_value in the validation crate but not in this PR.

rust/validation/src/validatable/mod.rs
```diff
@@ -249,13 +255,38 @@ pub trait ValidatableMappingPair<'a> {
...
+    ///
+    /// This must return a stable, human-readable key string even when
+    /// [`schema_key`](Self::schema_key) returns `None`, so validation can still
+    /// report a useful path for unexpected or ignored keys.
+    fn display_key(&self) -> Cow<'a, str>;
 
     /// Get the mapped value.
     fn value(&self) -> &'a Self::Value;
 
+    /// Build a coerced mapping entry from this pair's original key and a
```
gmuloc, line 291:
```suggestion
    /// Build a coerced mapping entry from this pair original key and a
```

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -331,32 +296,62 @@ impl<'a, 'input: 'a> Iterator for NodeMappingIter<'a, 'input> {
...
+    fn schema_key(&self) -> Option<Cow<'a, str>> {
+        schema_key_to_string(&self.pair.key)
     }
 
+    /// Return a path-friendly representation for diagnostics.
+    ///
+    /// This is deliberately broader than `schema_key` so numeric, boolean,
+    /// null, and complex YAML keys can still produce a useful validation path.
+    fn display_key(&self) -> Cow<'a, str> {
+        display_key_to_string(&self.pair.key)
```
gmuloc: wondering if the separate helper is necessary given that today it is only used here it seems. (though I may be missing future usage)

rust/validation/src/validatable/yaml_parser_impl.rs
```diff
@@ -331,32 +296,62 @@ impl<'a, 'input: 'a> Iterator for NodeMappingIter<'a, 'input> {
...
     fn value(&self) -> &'a Self::Value {
         &self.pair.value
     }
 
+    /// Build the YAML mapping pair for coerced output.
+    ///
+    /// String keys are rebuilt cheaply with the original key span; non-string
+    /// keys are cloned to keep their YAML value type, and the original pair
+    /// span is reused.
+    fn coerced_item(
```
gmuloc, line 330: I understand that this may be for performance and because in 99% cases we expect string keys but this makes a bit of inconsistency between string keys and non string keys. Maybe at least worth stating that other metadata is technically not needed today for string keys and thats why we chose the cheap option

rust/validation/src/validation/dict.rs
```diff
@@ -88,16 +88,15 @@ fn validate_keys<'a, M: ValidatableMapping<'a>>(
     schema: &Dict,
     input: &M,
     ctx: &mut Context,
-) -> Option<Vec<(String, <M::Value as ValidatableValue>::Coerced)>> {
+) -> Option<Vec<<M::Value as ValidatableValue>::CoercedMappingItem>> {
     let mut coerced_items = ctx.configuration.return_coerced_data.then(Vec::new);
 
     let Some(keys) = &schema.keys else {
         // No schema keys - preserve all input as-is when coercing
         if let Some(ref mut items) = coerced_items {
```
gmuloc, line 96: this is never tested so we could add the following test:


```rust
  #[test]
  fn validate_no_schema_keys_preserves_input_when_coercing() {
      let schema = Dict::default();
      let input = serde_json::json!({
          "foo": 123,
          "nested": {
              "bar": true
          }
      });

      let store = get_test_store();
      let configuration = Configuration {
          return_coerced_data: true,
          ..Default::default()
      };
      let mut ctx = Context::new(&store, Some(&configuration));

      let coerced = schema.validate(&input, &mut ctx);

      assert!(ctx.result.errors.is_empty());
      assert_eq!(coerced, Some(input));
  }
```

rust/validation/src/validation/dict.rs
gmuloc: we can add to this a configuration with return_coerced_data: true to make sure that removed keys are not coerced (need to assert then on the coerced value output from schema.validate

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: keys are no longer coerced.

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: emitting a new OutOfRange violation now.

rust/validation/src/validation/store.rs
ClausHolbechArista: That would just test the yaml parser. All these would be null as discussed.

rust/validation/src/validatable/mod.rs
ClausHolbechArista: ?

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: let's keep it for now. It is performant enough and we should rather work on bigger things like not having to return all data as coerced etc.

rust/validation/src/validation/store.rs
ClausHolbechArista: added

rust/validation/src/validation/store.rs
ClausHolbechArista: enhanced this test to cover two documents.

rust/validation/src/validatable/mod.rs
gmuloc: me and my sad English teacher apparently

rust/validation/src/validation/dict.rs
```diff
@@ -117,66 +116,80 @@ fn validate_keys<'a, M: ValidatableMapping<'a>>(
...
-                        .unwrap_or_else(|| input_value.clone_to_coerced());
-                    items.push((input_key_str.to_owned(), coerced));
-                } else {
-                    let _ = key_schema.validate(input_value, ctx);
+                false // Already handled
+            } else if let Some(dynamic_keys_infos) = &dynamic_keys_infos
+                && let Some(dynamic_key_info) = dynamic_keys_infos.get(input_schema_key)
+            {
+                let key_schema = dynamic_key_info.schema;
+                if !check_deprecation(input_schema_key, key_schema, key_span, input, ctx) {
```
gmuloc, line 150: we don't have tests for deprecated dynamic keys today and we will soon need to do so as we remove them so worth adding it for now:


```rust
  #[test]
  fn validate_dynamic_key_deprecated_ok() {
      let schema: Dict = Dict::deserialize(serde_json::json!({
          "keys": {
              "my_dynamic_keys": {
                  "type": "list",
                  "items": {
                      "type": "dict",
                      "keys": {
                          "key": {
                              "type": "str"
                          }
                      }
                  }
              }
          },
          "dynamic_keys": {
              "my_dynamic_keys.key": {
                  "type": "int",
                  "deprecation": {
                      "warning": true,
                      "remove_in_version": "1.2.3"
                  }
              }
          },
          "allow_other_keys": true
      }))
      .unwrap();

      let input = serde_json::json!({
          "my_dynamic_keys": [{"key": "dynkey1"}],
          "dynkey1": 5
      });

      let store = get_test_store();
      let mut ctx = Context::new(&store, None);
      let _ = schema.validate(&input, &mut ctx);

      assert!(ctx.result.errors.is_empty());
      assert_eq!(
          ctx.result.warnings,
          vec![Feedback {
              path: vec!["dynkey1".into()].into(),
              span: None,
              issue: WarningIssue::Deprecated(Deprecated {
                  path: vec!["dynkey1".into()].into(),
                  replacement: None.into(),
                  version: Some("1.2.3".into()).into(),
                  url: None.into(),
              })
          }]
      );
  }
```

ClausHolbechArista: no, this will be changed in a later pr where all keys (dynamic, prefix or static) will be handles with the same logic.

ClausHolbechArista: this is general validation coverage.

rust/validation/src/validation/dict.rs
ClausHolbechArista: why do we care? removed keys will block further execution.

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: it is not needed for any key. You said we didn't care in your previous comments but you seem to care :) I will clone all of it now so we can move on.

gmuloc (approved):
LGTM other than the nits about coverage that can be tackled in another PR

rust/validation/src/validatable/yaml_parser_impl.rs
ClausHolbechArista: removed helper

rust/validation/src/validation/dict.rs
ClausHolbechArista: added

rust/validation/src/validation/dict.rs
ClausHolbechArista: added test

rust/validation/src/validation/dict.rs
ClausHolbechArista: added test

PR #132, opened 2026-05-19
CI: Use our own pre-commit hooks repo

ClausHolbechArista:
- Use pre-commit in CI workflow instead of pre-commit.ci.
- Use our own hooks repo for cargo based hooks.
- Use autofix.ci for committing back fixes.

Tested and confirmed the correct fixes are committed back here: https://github.com/aristanetworks/pyavd-utils/pull/142/commits/9c7ec0bf443a43f2ee5b7a84a0ab67061f081a44

### DIFF
--- a/.github/workflows/pull-request-autofix.yml
+++ b/.github/workflows/pull-request-autofix.yml
@@ -35,6 +35,7 @@ jobs:
 
       - name: Run pre-commit
         uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        continue-on-error: true
         with:
           extra_args: --all-files
         # Allowing cargo installs to be unoptimized for install speed.
--- a/.github/workflows/pull-request-autofix.yml
+++ b/.github/workflows/pull-request-autofix.yml
@@ -7,6 +7,10 @@ name: autofix.ci
     branches:
       - main
 
+concurrency:
+  group: autofix-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
@@ -16,10 +20,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.12"
 
@@ -28,7 +34,7 @@ jobs:
         run: rustup toolchain install nightly --component rustfmt --no-self-update
 
       - name: Run pre-commit
-        uses: pre-commit/action@v3.0.1
+        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
         with:
           extra_args: --all-files
         # Allowing cargo installs to be unoptimized for install speed.
--- a/.github/workflows/pull-request-autofix.yml
+++ b/.github/workflows/pull-request-autofix.yml
@@ -45,4 +45,4 @@ jobs:
           CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 256
 
       - name: Commit back any git diff
-        uses: autofix-ci/action@c5b2d67aa2274e7b5a18224e8171550871fc7e4a
+        uses: autofix-ci/action@c5b2d67aa2274e7b5a18224e8171550871fc7e4a # v1.3.4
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,41 @@
+---
+name: autofix.ci
+
+"on":
+  pull_request:
+  push:
+    branches:
+      - main
```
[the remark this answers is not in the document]
ClausHolbechArista: done

.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      # Needed until we move to prek and get proper support for toolchain profile - preferably per hook.
+      - name: Install Rust nightly formatter
+        run: rustup toolchain install nightly --component rustfmt --no-self-update
+
+      - name: Run pre-commit
+        uses: pre-commit/action@v3.0.1
```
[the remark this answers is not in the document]
ClausHolbechArista: done

.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,41 @@
...
+permissions:
+  contents: read
+
+jobs:
+  autofix:
+    name: Run pre-commit and commit back any changes
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v6
```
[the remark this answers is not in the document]
ClausHolbechArista: done

.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,41 @@
...
+          python-version: "3.12"
+
+      # Needed until we move to prek and get proper support for toolchain profile - preferably per hook.
+      - name: Install Rust nightly formatter
+        run: rustup toolchain install nightly --component rustfmt --no-self-update
+
+      - name: Run pre-commit
+        uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: --all-files
```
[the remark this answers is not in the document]
ClausHolbechArista: done

.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Run pre-commit
+        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        continue-on-error: true
+        with:
+          extra_args: --all-files
+        # Allowing cargo installs to be unoptimized for install speed.
+        env:
+          CARGO_PROFILE_RELEASE_OPT_LEVEL: 0
+          CARGO_PROFILE_RELEASE_LTO: false
+          CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 256
```
gmuloc, line 45: not blocking but why do we need this?

.github/workflows/pull-request-autofix.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Commit back any git diff
+        uses: autofix-ci/action@c5b2d67aa2274e7b5a18224e8171550871fc7e4a # v1.3.4
```
gmuloc, line 48: don't we want to show fail-fast: true (even if it is the default)? to be clear in our intent?

gmuloc (commented):
Image

this could explode very soon no?

.github/workflows/pull-request-autofix.yml
ClausHolbechArista: as it says in the comment above :) 

# Allowing cargo installs to be unoptimized for install speed.

this will install pre-compiled binaries if available and lowers installation from ~4 min to ~1 min.

.github/workflows/pull-request-autofix.yml
ClausHolbechArista: there is no matrix here? How would fail-fast have an impact here?

.github/workflows/pull-request-autofix.yml
gmuloc: well it feels weird say we want unoptimized :p but ok I understand now

gmuloc:
> Image
> this could explode very soon no?

Apparently it will work

.github/workflows/pull-request-autofix.yml
gmuloc: 1/ it says 

```
inputs:
  fail-fast:
    description: 'Cancel all other workflows associated with a commit when fixing it.'
```

would it not cut - our other workflows (pull-request-management)? 

2/ this is the default so anyways it applies just saying we should be stating it explicitely

ClausHolbechArista: that is already handled with the concurrency rule I think. Anyway no need to bloat it here I think.

ClausHolbechArista: so we think this will cancel _other_ workflows for the same commit to save on CI time.

gmuloc (approved):

PR #133, opened 2026-05-21
Feat(rust): Add support for dynamic key overrides

ClausHolbechArista:
## Change Summary

Add support for dynamic key overrides

## Component(s) name

rust

## Proposed changes

For use by the LSP-server, we need to be able to override which keys are dynamic keys and which schema they are mapped to. This is what enables us to interpret comments like below and assign the correct schema, even when the `network_services_keys` are in a different file.

```yaml
mytenants: # dynamic_key: network_services
  ...
```

Also updating some helper functions on avdschema to make hover and completion work for the same.

This is not used for regular runtime validation.

## How to test

- Added tests coverage
- Used in prototype LSP-server and works.

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

* **New Features**
  * Support for dynamic key overrides in dictionary schemas to map concrete keys to schema paths.
  * Configuration can carry optional dynamic-key overrides for validation.
  * Static keys take precedence over dynamically-derived keys and overrides.

* **Bug Fixes**
  * Removed dynamic-key entries are now excluded across the entire dynamic-keys map, not just at the start.

* **Tests**
  * Tests expanded to cover override resolution and precedence behaviors.

### DIFF
--- a/rust/avdschema/src/schema/dict.rs
+++ b/rust/avdschema/src/schema/dict.rs
@@ -87,7 +87,7 @@ impl<'a> Dict {
         self.dynamic_keys.as_ref().map(|dynamic_keys| {
             let mut resolved_dynamic_keys: OrderMap<String, DynamicKeyInfo<'a>> = dynamic_keys
                 .iter()
-                .skip_while(|(_, dynamic_key_schema)| dynamic_key_schema.is_removed())
+                .filter(|(_, dynamic_key_schema)| !dynamic_key_schema.is_removed())
                 .flat_map(|(dynamic_key_path, dynamic_key_schema)| {
                     self.get_dynamic_key_values(dynamic_key_path, dict)
                         .into_iter()
@@ -148,7 +148,7 @@ impl<'a> Dict {
         self.dynamic_keys.as_ref().map(|dynamic_keys| {
             dynamic_keys
                 .iter()
-                .skip_while(|(_, dynamic_key_schema)| dynamic_key_schema.is_removed())
+                .filter(|(_, dynamic_key_schema)| !dynamic_key_schema.is_removed())
                 .flat_map(|(dynamic_key_path, _)| {
                     dynamic_key_path
                         .split('.')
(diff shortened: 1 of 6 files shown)


rust/validation/src/context.rs
```diff
@@ -170,6 +173,11 @@ pub struct Configuration {
     /// When validating avd_design, emit warnings for top-level keys that exist in eos_config
     /// but not in avd_design.
     pub warn_eos_config_keys: bool,
+    /// Optional caller-supplied dynamic key overrides keyed by concrete input key.
+    /// The override value is the schema dynamic-key path that should be used for that key.
+    /// This is used by the LSP when interpreting # comments on keys.
+    /// Stored behind Arc so cloning Configuration for each new Context stays cheap.
+    pub dynamic_key_overrides: Option<Arc<DynamicKeyOverrides>>,
```
gmuloc: I think we said the order does not matter so we can maybe put this one in alphabetical order like the rest?

rust/validation/src/validation/dict.rs
```diff
@@ -575,6 +581,68 @@ mod tests {
...
+                Int {
+                    max: Some(10),
+                    ..Default::default()
+                }
+                .into(),
+            )])),
+            allow_other_keys: Some(true),
+            ..Default::default()
+        };
+        let input = serde_json::json!({ "dynkey1": 5 });
```
gmuloc, line 594: if you change this to 11 the test won't fail because of the ealry return line 94 where we don't validate if there is no keys in the schema (which is the case above)

so the test does not test what it thinks it does.

rust/validation/src/validation/dict.rs
gmuloc, line 94: essentially today we wont validate dynamic_keys only schema (which was the case before)

rust/avdschema/src/schema/dict.rs
```diff
@@ -366,6 +426,186 @@ mod tests {
...
+            ])),
+            ..Default::default()
+        };
+        let value: Value = json!({
+            "kept_list": ["kept"],
+            "removed_list": ["removed"],
+            "later_list": ["later"],
+        });
+
+        let result = dict_schema.get_dynamic_keys(value.as_object().unwrap(), None);
```
gmuloc: you could add an override here:

```
  let overrides = DynamicKeyOverrides::from_iter([(
      "removed_override".to_string(),
      "removed_list".to_string(),
  )]);
```

that would check that overrides for removed dynamic keys are ignored (exercising line 108/109)

ClausHolbechArista: Is this the line you were thinking of?

rust/validation/src/context.rs
ClausHolbechArista: this was authored long before we had that talk.

ClausHolbechArista: reordered

rust/validation/src/validation/dict.rs
ClausHolbechArista: fixed. Inserted keys and verified that it failed with an invalid value.

rust/validation/src/validation/dict.rs
gmuloc: ok for now - this is an ok limitation

rust/avdschema/src/schema/dict.rs
ClausHolbechArista: added

gmuloc (approved):

PR #135, opened 2026-05-22
CI: Enable workspace lint rules on every crate

gmuloc:
For now add allows that needs to be looked at afterwards

## Summary by CodeRabbit

* **Chores**
  * Applied workspace lint configuration across Rust crates for standardized code quality checks.
  * Refactored internal code patterns and string handling for consistency.
  * Enhanced documentation clarity with improved formatting and descriptions.
  * Optimized module visibility to better encapsulate internal implementation details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aristanetworks/pyavd-utils/pull/135?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #136, opened 2026-05-22
CI: Remove inline tests from coverage for more accurate indicators

gmuloc:
Right now our coverage numbers are also counting the inline tests. Using coverage nightly to fix this.

NOTE: this PR edits to Cargo.toml files can be changed once we align linting across all crates


gmuloc:
@ClausHolbechArista so this is a more accurate view of our coverage

rust/avdschema/Cargo.toml
```diff
@@ -26,3 +26,6 @@ xz2 = { version = "0.1.7", optional = true }
 [features]
 dump_load_files = ["dep:walkdir"]
 xz2 = ["dump_load_files", "dep:xz2"]
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ["cfg(coverage_nightly)"] }
```
ClausHolbechArista, line 31: this should be taken from the workspace

rust/avdschema/src/inherit.rs
```diff
@@ -202,6 +202,7 @@ impl Inherit for AnySchema {
 }
 
 #[cfg(test)]
+#[cfg_attr(coverage_nightly, coverage(off))]
```
ClausHolbechArista, line 205: there must be some global knob for this. We will not remember to add this.

rust/avdschema/Cargo.toml
gmuloc: we can't because we have been bad cituizens:
https://github.com/aristanetworks/pyavd-utils/pull/135

gmuloc:
As discussed - closing this for now. We will accept that we are having coverage on tests. Remembering to add all the instrumentation is a burden we are not ready to take on.
So we will know that currently, our coverage is skewed.

PR #159, opened 2026-06-11
Refactor(rust): Rearrance code for dynamic keys

ClausHolbechArista:
To cater for more options to resolve schema keys the code handling dict keys is being rearranged and abstracted into a DictKeyMatch result. Static and DynamicKeys are the only variants right now but an example of extending can be seen in the prefix-keys PR.
This would also be where the generated _profiles keys could get injected.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic dictionary key support with schema-derived (dynamic) keys and optional override mappings to control key resolution.
  * Updated schema key resolution to properly handle static vs dynamic keys during `$ref` generation.

* **Improvements**
  * Enhanced dictionary key validation to use unified static/dynamic/unknown matching, including better handling of non-string keys.
  * Refined “allow other keys” behavior and EOS config warnings while preserving original key/value pairs when passthrough is enabled.
  * Added a unified `keys()` iterator for mapping data used by schema/key resolution.


ClausHolbechArista:
@coderabbitai review

rust/avdschema/src/schema/dict/dynamic_keys.rs
```diff
@@ -0,0 +1,57 @@
...
+            && let Some(schema) = static_keys.get(key)
+        {
+            return DictKeyMatch::Static(schema);
+        }
+
+        if let Some(dynamic_keys) = &self.dynamic_keys
+            && let Some(dynamic_key_info) = dynamic_keys.get(key)
+        {
+            return DictKeyMatch::Dynamic(dynamic_key_info.clone());
+        }
```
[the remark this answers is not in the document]
ClausHolbechArista: fixed. Also fixed the same in schema_from_path

rust/avdschema/src/utils/schema_from_path.rs
```diff
@@ -2,65 +2,73 @@
...
             },
             None => return schema_ref,
         }
-        for step in path {
-            if step.parse::<usize>().is_ok() {
-                schema_ref.push_str("/items");
-            } else {
-                schema_ref.push_str("/keys/");
-                schema_ref.push_str(step);
+        let mut append_path = |steps: &mut dyn Iterator<Item = &String>| {
```
gmuloc: I don't understand why we are adding the mut here? the previous code (inner for loop) still works. is this some clippy thing?

gmuloc (approved):
one nit as I think we are introducing unneeded code but maybe I am missing why. Otherwise LGTM

rust/avdschema/src/utils/schema_from_path.rs
ClausHolbechArista: Good catch. The mut is only needed because the new closure mutably captures schema_ref, so the closure is FnMut. But the closure is not actually buying us anything here, and the old inline loop still works. I’ll revert this part back to the simpler loop.

gmuloc (approved):

PR #190, opened 2026-07-21
Refactor: Move Python bindings into shared module

gmuloc:
## Change Summary

In order to prepare for moving schema-merging capabilities to pyavd-utils, we require the store to be available across python modules.
The solution is to make one big `so` file

keeping the previously exposed public API

## Component(s) name

`validation`, `passwords`, `rust`

## Proposed changes

- Replace separate pyvalidation and pypasswords extension modules with one shared pyavd_utils._bindings extension.
- Add Python wrapper modules for the public API:
    - pyavd_utils.validation
    - pyavd_utils.passwords
    - pyavd_utils.schema_store
- Move schema-store initialization into the shared Rust binding crate using a single OnceLock<Store>.
- Keep `pyavd_utils.validation.init_store_from_file()` as a deprecated compatibility wrapper.
- Preserve password features behind the existing Rust feature flags.
- Preserve the previous validation performance benchmark under rust/python_bindings.
- Use `pyo3_build_config::add_extension_module_link_args()` for the single extension module. With PyO3/pyo3-build-config 0.29.0, this is required for macOS extension-module linking and is a no-op on normal non-macOS targets hence the simplification

## How to test

Tested locally on macOS arm64 with Python 3.11.11.

pre-commit passes
all tests are passing

### Repository Checklist

- [x] I have read the CONTRIBUTING (https://avd.arista.com/devel/docs/contribution/overview.html) document.
- [x] I have updated testing accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added unified Python bindings for password utilities, schema-store initialization, and validation.
  * Added support for validating JSON against ad-hoc schemas.
  * Exposed validation results, violations, deprecations, and ignored configuration keys through the public API.
  * Added validated-data retrieval when input passes validation.
  * Updated password verification to return a boolean result.

* **Bug Fixes**
  * Improved error reporting for invalid JSON, schemas, passwords, and encryption inputs.
  * Corrected public API documentation and type annotations.


pyavd_utils/validation.py
```diff
@@ -0,0 +1,46 @@
...
+    Configuration,
+    Deprecation,
+    IgnoredEosConfigKey,
+    ValidatedDataResult,
+    ValidationResult,
+    Violation,
+    get_validated_data,
+    validate_json,
+    validate_json_with_adhoc_schema,
+)
```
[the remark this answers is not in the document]
gmuloc: so I had this originally but then I need to maintain the same pyi multiple times? or cna I just maybe have one pyi and it will show up in the passwords reimoprt?

pyavd_utils/passwords.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2025-2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from ._bindings import cbc_decrypt, cbc_encrypt, cbc_verify, sha512_crypt, simple_7_decrypt, simple_7_encrypt
```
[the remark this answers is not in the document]
gmuloc: should be addressed now?

pyavd_utils/validation.py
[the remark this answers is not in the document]
gmuloc: done

rust/python-bindings/src/tests/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+
+fn init_test_store(py: pyo3::Python<'_>) {
+    assert!(STORE.get().is_none(), "Already set");
+    let module = py.import("_bindings").unwrap();
+    let args = ();
+    let kwargs = pyo3::types::PyDict::new(py);
+    let file = py.detach(test_schema_store::get_store_gz_path);
+    kwargs.set_item("file", file).unwrap();
+    let _ = module.call_method("init_store_from_file", args, Some(&kwargs));
+}
```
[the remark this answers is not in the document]
gmuloc: done

rust/python-bindings/benches/performance.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2025-2026 Arista Networks, Inc.
+// Use of this source code is governed by the Apache License 2.0
+// that can be found in the LICENSE file.
+//! Criterion benchmarks.
+#![allow(
+    missing_docs,
+    reason = "criterion_group generates an undocumented benchmark entrypoint function"
+)]
```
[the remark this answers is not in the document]
gmuloc: please double check this one, it should be addressed

pyavd_utils/THIRD_PARTY_LICENSES.txt
```diff
@@ -10,8 +10,7 @@ Apache License 2.0
 Used by:
   - avdschema 0.0.7
   - passwords 0.0.7
-  - pypasswords 0.0.7
-  - pyvalidation 0.0.7
+  - python_bindings 0.0.7
```
ClausHolbechArista: use dash for crate names

pyavd_utils/_bindings/__init__.pyi
ClausHolbechArista, line 1: I hoped we could keep the python side split into modules, so _validation and _password and now _schema_store. Is that not possible?

gmuloc: we need one so file from which we import so we can have the same store later so I don't think we can do this from what I have read.
There would be an option to create a dynamic library owning the store and create three so files linking to this library but there are cons like: cross platform support to load the library, tooling handling (auditwheel), ...

ClausHolbechArista: ack

pyavd_utils/schema_store.pyi
ClausHolbechArista, line 1: this file should not be needed.

pyavd_utils/validation.pyi
```diff
@@ -1,143 +1,30 @@
...
-    Args:
-        data_as_json: Structured data dumped as JSON.
-        schema_as_json: A fully resolved schema dumped as JSON.
-        configuration: Optional configuration for validation behavior.
-
-    Returns:
-        ValidationResult holding lists of violations and deprecations.
-    """
+from ._bindings import (
+    Configuration as Configuration,
```
ClausHolbechArista: why not just `from ._bindings import Configuration`?

pyavd_utils/schema_store.pyi
gmuloc: just tested and indeed they are not

rust/python-bindings/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+use pyo3::Bound;
+use pyo3::PyResult;
+use pyo3::pymodule;
+use pyo3::types::PyModule;
+
+mod passwords;
+mod schema_store;
+mod validation;
+
+#[pymodule(gil_used = false)]
```
ClausHolbechArista: is this safe for everything? We did not have this before.

rust/python-bindings/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+#[pyo3(name = "_bindings")]
+pub fn bindings(module: &Bound<'_, PyModule>) -> PyResult<()> {
```
ClausHolbechArista: I prefer to use the inline model where we do a bindings module and define/reexport the things below. We can do it with proper module files here and just put the pymodule on the mod command. Let's discuss live.

rust/python-bindings/src/tests/passwords/test_cbc.rs
```diff
@@ -2,32 +2,32 @@
 // Use of this source code is governed by the Apache License 2.0
 // that can be found in the LICENSE file.
 
-use super::*;
+use pyo3::types::PyAnyMethods as _;
+
+use super::super::setup_python;
```
ClausHolbechArista: could we use crate::tests:setup_python?

rust/test_schema_store/Cargo.toml
```diff
@@ -2,7 +2,10 @@
 name = "test_schema_store"
 edition = "2024"
 
+[workspace]
```
ClausHolbechArista: ?

rust/test_schema_store/Cargo.toml
```diff
@@ -2,7 +2,10 @@
...
+
 [dependencies]
+pyo3 = { version = "0.29.0", features = ["abi3-py310"] }
```
ClausHolbechArista: this will bring pyo3 into all tests of validation etc. Let's gate this behind a python-bindings feature if it is really needed here. I would prefer not to have any pyo3 here.

rust/test_schema_store/src/lib.rs
```diff
@@ -33,3 +38,25 @@ fn _get_store_gz_path() -> PathBuf {
...
+            let module = py.import("_bindings").unwrap();
+            let kwargs = PyDict::new(py);
+            let file = py.detach(get_store_gz_path);
+            kwargs.set_item("file", file).unwrap();
+            module
+                .call_method("init_store_from_file", (), Some(&kwargs))
+                .unwrap();
+        });
+    });
+}
```
ClausHolbechArista: IMO this should live in the python bindings crate. What am I missing?

tests/schema_store/test_schema_store.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025-2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
```
gmuloc, line 4: this shoul go under its own folder not under validation

gmuloc: and the fixture need to be moved unde rtests

rust/python-bindings/src/lib.rs
gmuloc: so turns out it is default since 0.28 https://pyo3.rs/main/migration?utm_source=openai#default-to-supporting-free-threaded-python so I will just remove this

rust/python-bindings/src/tests/passwords/test_cbc.rs
gmuloc: done

rust/python-bindings/src/lib.rs
gmuloc: tried something

rust/test_schema_store/Cargo.toml
gmuloc: this one is because of my setup locally working on it

rust/python-bindings/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+use pyo3::pymodule;
+use pyo3::types::PyModule;
+
+mod passwords;
+mod schema_store;
+mod validation;
+
+#[pymodule]
+#[pyo3(name = "_bindings")]
+pub mod bindings {
```
gmuloc:
```suggestion
pub mod _bindings {
```

rust/test_schema_store/Cargo.toml
gmuloc: removed and du0plication ok

pyavd_utils/passwords.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from ._bindings import passwords as _passwords
```
gmuloc: try importing funcitons directly

pyavd_utils/_bindings.pyi
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from typing import Any
```
gmuloc: _bindings folder with only pyi

pyavd_utils/passwords.py
gmuloc: does not work

ClausHolbechArista (approved):

PR #197, opened 2026-07-28
Feat: Add schema list primary key helper

gmuloc:
## Change Summary

Add a schema-store helper to return the primary key for list schemas at a given data path.

## Related Issue(s)

N/A

## Component(s) name

`python`

## Proposed changes

- Add `pyavd_utils.schema_store.get_list_primary_key(schema_name, data_path)`.
- Expose the Rust schema path resolver errors needed by the Python binding.
- Resolve the schema for a data path, return the list schema `primary_key` when present, and return `None` for unknown paths, non-list paths, or unresolved schema-walk paths.
- Document the current limitation: user-defined dynamic root keys are not resolved because the helper does not accept input data or dynamic-key overrides.
- Add Python and Rust binding tests for

## How to test

Tested with:

```shell
cargo fmt --check
cargo check -p python-bindings
uv run pytest tests/schema_store/test_schema_store.py
```
### User Checklist

- Add dynamic key support & tests

### Repository Checklist

- [x] I have read the CONTRIBUTING (https://avd.arista.com/devel/docs/contribution/overview.html) document.

## Summary by CodeRabbit

* **New Features**
  * Added `get_list_primary_key` to retrieve a list schema’s primary key from a data path for `eos_config`.
  * Returns `None` when the target is not a list or cannot be resolved; unsupported schema names produce an error.
  * Available through the Python schema store interface.
* **Refactor**
  * Improved consistency of schema name and alias resolution.
* **Tests**
  * Expanded coverage for top-level, nested, non-list, unknown, and unsupported-schema cases.

### DIFF
--- a/pyavd_utils/_bindings/_schema_store.pyi
+++ b/pyavd_utils/_bindings/_schema_store.pyi
@@ -11,8 +11,8 @@ def get_list_primary_key(schema_name: str, data_path: list[str]) -> str | None:
 
     Limitation:
         This only supports the EOS config schema for now, since other AVD schemas can use
-        dynamic keys which are not supported by this helper yet. Supported schema names are
-        "eos_config" and the "eos_cli_config_gen" alias.
+        dynamic keys which are not supported by this helper yet. The only supported schema
+        name is "eos_config".
 
     Args:
         schema_name: The name of the schema to inspect.
@@ -20,7 +20,8 @@ def get_list_primary_key(schema_name: str, data_path: list[str]) -> str | None:
 
     Raises:
         RuntimeError: If the shared schema store has not been initialized, if the schema name is
-            not supported, or if schema resolution fails.
+            not supported, or if schema resolution fails for reasons other than an unresolved
+            schema path. Schema walk failures, such as unresolved nested-list paths, return None.
     """
 
 def init_store_from_file(file: Path) -> None:
--- a/pyavd_utils/_bindings/_schema_store.pyi
+++ b/pyavd_utils/_bindings/_schema_store.pyi
@@ -5,7 +5,9 @@
 # ruff: noqa: PYI021
 from pathlib import Path
 
-def get_list_primary_key(schema_name: str, data_path: list[str]) -> str | None:
+from typing import Literal
+
+def get_list_primary_key(schema_name: Literal["eos_config"], data_path: list[str]) -> str | None:
     """
     Return the primary key for a list schema at the given data path.
 
--- a/pyavd_utils/_bindings/_schema_store.pyi
+++ b/pyavd_utils/_bindings/_schema_store.pyi
@@ -5,6 +5,24 @@
 # ruff: noqa: PYI021
 from pathlib import Path
 
+def get_list_primary_key(schema_name: str, data_path: list[str]) -> str | None:
+    """
+    Return the primary key for a list schema at the given data path.
+
+    Limitation:
+        This only resolves static schema paths and dynamic root keys that can be
+        inferred from schema defaults. User-defined dynamic root keys are not
+        resolved because this helper does not accept input data or dynamic-key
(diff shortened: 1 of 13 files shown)


rust/python-bindings/src/schema_store/mod.rs
```diff
@@ -52,4 +56,36 @@ pub(crate) mod _schema_store {
             )
         }).inspect(|()| info!("Initialized the schema store from file."))
     }
+
+    #[pyfunction]
```
ClausHolbechArista, line 59: I think we should put this into the avd_schema crate instead and just have a thin wrapper here.

rust/avdschema/src/store.rs
```diff
@@ -32,23 +32,46 @@ impl Store {
         schema_names
     }
 
-    pub fn get(&self, schema_name: &str) -> Result<&AnySchema, SchemaStoreError> {
+    fn fetch<'a, 'b>(
```
ClausHolbechArista: this change affects lsp too, but I will manage...

ClausHolbechArista: ah no, saw the get below now.

rust/avdschema/src/utils/schema_from_path.rs
```diff
@@ -115,6 +115,7 @@ pub enum GetSchemaFromPathError {
     StoreError(SchemaStoreError),
     Keys(SchemaKeysError),
     Resolve(SchemaResolverError),
+    UnsupportedSchemaName(String),
```
ClausHolbechArista: isn't this already covered by the StoreError or do you want this to be specific?

rust/avdschema/src/utils/schema_from_path.rs
```diff
@@ -157,9 +158,43 @@ pub fn get_schema_from_path<'store, 'value>(
...
+///
+/// This helper only supports the `eos_config` schema for now, since other
+/// schemas can use dynamic keys which are not supported here.
+pub fn get_list_primary_key(
+    schema_name: &str,
+    store: &Store,
+    data_path: &[String],
+) -> Result<Option<String>, GetSchemaFromPathError> {
+    let canonical_schema_name = store.canonical_schema_name(schema_name)?;
+    if !matches!(canonical_schema_name, "eos_config" | "eos_cli_config_gen") {
```
ClausHolbechArista: are you not normalizing the canonical_schema_name to eos_config?

rust/avdschema/src/utils/schema_from_path.rs
ClausHolbechArista: strictly speaking this should likely be in it's own enum wrapping the GetSchemaFromPathError enum and then adding this extra one. Since now someone calling the get_schema_from_path will have to match on this that will never happen for them.

rust/avdschema/src/utils/schema_from_path.rs
```diff
@@ -169,6 +204,52 @@ mod tests {
     use crate::str::Str;
     use crate::utils::test_utils::get_test_store;
 
+    fn get_list_primary_key_test_store() -> Store {
```
ClausHolbechArista: I think we already have some of these in the test_utils? Otherwise add it there.

pyavd_utils/_bindings/_schema_store.pyi
```diff
@@ -5,6 +5,25 @@
 # ruff: noqa: PYI021
 from pathlib import Path
 
+def get_list_primary_key(schema_name: str, data_path: list[str]) -> str | None:
```
ClausHolbechArista:
```suggestion
from typing import Literal

def get_list_primary_key(schema_name: Literal["eos_config"], data_path: list[str]) -> str | None:
```

ClausHolbechArista (approved):

